Issue #8, opened 2023-01-23
Add option to reset user limit at different points in time


The-Daishogun:
Hi, thank you for the amazing work you're doing 🙌🏼 

how do you feel about adding an option to make user bandwith limit rest weekly or monthly or even yearly?

i can get started on the pull request if you feel this would be a good adition

gozarman:
Hi @The-Daishogun, 
First of all, thanks for helping to improve Marzban.

This feature seems cool, feel free to make a PR.

Just before implementing, could you pls tell us more about how you gonna add this feature? (Which APIs are going to be changed) and how it's going to be implemented on the UI? 

I would appreciate it if you use the [Feature request template](https://github.com/Gozargah/Marzban/issues/new/choose) for future requests.

Issue #11, opened 2023-01-25
Options must be added for user management


SaintShit:
A TODO list of features requested on user modifying methods

- [x] Option to active / deactive user temporarily
- [x] Option to reset user's usage
<strike>- [ ] Option to import / export backup of users</strike>
<strike>- [ ] XTLS support for VLESS (flow), specially for `xtls-rprx-vision`</strike>
<strike>- [ ] Option to add note on users</strike>

PR #20, opened 2023-02-01
Feature/reset user data usage

The-Daishogun:
Closes issue #8 and addresses one TODO from #11.


SaintShit:
Thanks for your neat code! I'm gonna have a review

app/db/crud.py
```diff
@@ -34,6 +35,11 @@ def get_users(db: Session,
             query = query.filter(User.status.in_(status))
         else:
             query = query.filter(User.status == status)
+    if reset_strategy:
+        if isinstance(status, list):
```
SaintShit, line 39: `status` must change with `reset_strategy`

app/jobs/reset_user_data_usage.py
```diff
@@ -0,0 +1,32 @@
...
+
+reset_strategy_to_days = {
+    UserDataLimitResetStrategy.day.value :1,
+    UserDataLimitResetStrategy.week.value: 7,
+    UserDataLimitResetStrategy.month.value : 30,
+    UserDataLimitResetStrategy.year.value : 365,
+}
+
+def reset_user_data_usage():
+    now = datetime.utcnow().timestamp()
```
SaintShit, line 15: you must remove `.timestamp()` to prevent error below in following lines
`TypeError: unsupported operand type(s) for -: 'float' and 'datetime.datetime'`

app/jobs/reset_user_data_usage.py
```diff
@@ -0,0 +1,32 @@
...
+        num_days_to_reset = reset_strategy_to_days[user.data_limit_reset_strategy]
+        if not  (now - last_reset_time).days >= num_days_to_reset:
+            continue
+        
+        crud.reset_user_data_usage(db, user)
+
+        logger.info(f"User data usage reset for User \"{user.username}\"")
+
+
+scheduler.add_job(reset_user_data_usage, 'interval', hours=24)
```
SaintShit, line 32: How you feel about changing this to 1 hour or less? this 24 hour will be calculated from runtime. by decreasing the interval, you can increase the accuracy of the reset to be closer to 24 hours. this is suitable we need a 24 hour reset.
one another idea is to set this job to run at the end of every UTC day, therefore it can mean "daily" more.
what's your idea?

app/db/crud.py
```diff
@@ -19,7 +19,8 @@ def get_users(db: Session,
               limit: int = None,
               username: str = None,
               status: Union[UserStatus, list] = None,
-              admin: Admin = None):
+              admin: Admin = None,
+              reset_strategy: Union[UserDataLimitResetStrategy, list] = None) -> list[User]:
```
SaintShit, line 23: Please change `list[User]` to `typing.List[User]` to be compatible with Python3.8 too.

SaintShit (changes requested):

PR #23, opened 2023-02-06
Reset User data limit

parasite249:
Hi since #20 was closed for no apparent reason and i could really use the changes from it, i copied the changes in this PR.


SaintShit:
Good job, Thanks!

PR #27, opened 2023-02-07
add homepage

OxJ4ck:
add real word homepage.


SaintShit:
Hi there, thanks for your investigation it means a lot to us.
the changes you made is really interesting and helpful. but the webpage you added is almost 20MB. this is not suitable for a part which is not necessary much. I mean it cloud be better if we have just one file like nginx, caddy, or apache2. and let the user decide to what's gonna show up to users.

PR #30, opened 2023-02-08
Status header for Clash and Shadowrocket, sub update interval

reloadlife:
- Shows expire date, and usage data in application while using subscriptions
- added a 24h interval for subscription update (clash)


SaintShit:
Thanks 🤓

PR #31, opened 2023-02-08
Custom SNI Support on XRAY_HOSTS

reloadlife:
- Custom SNI in XRAY_HOSTS (format Name@ip@SNI) SNI is optional
- Multi SNI will work too: (e.g: Marz@10.10.10.1@example.com,example.org,exmaple.net)

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -117,7 +117,9 @@ class UserResponse(User):
                     link = ShareLink(f"{host['remark']} ({values['username']})",
                                      host['hostname'],
                                      proxy_type,
-                                     settings.dict())
+                                     settings.dict(),
+                                     sni=host["sni"]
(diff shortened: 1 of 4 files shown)


SaintShit:
Hi there, thanks for your contribution. have you tested it out? My tests didn't go well.
I've set xray hosts to
```bash
XRAY_HOSTS="Marz@10.10.10.1@example.com"
```
and the result was
```python
[{'remark': 'Marz@10.10.10.1', 'hostname': 'example.com', 'sni': None}]
```

Also I wonder how does the multi-SNI feature work, as I tested, clients such V2rayNG doesn't accept SNI such this: `example.com,example.org,exmaple.net` and I don't see any code to randomly pick one of them. could you explain it please?

PR #32, opened 2023-02-08
Fixed TelegramBot

reloadlife:
- Fixed an issue in telegram bot that /user <username> was not working due to message being too long when there are too many proxies available.
- Added /create to create users from telegram bot.


SaintShit:
#42

PR #34, opened 2023-02-13
show traffic & status information for public users in telegram .

bahram1249:
show traffic information for public users in telegram .

### DIFF
--- a/app/telegram/admin.py
+++ b/app/telegram/admin.py
@@ -1,6 +1,5 @@
-import math
+from app.utils.readable import readable_size
 from datetime import datetime
-
 from app import xray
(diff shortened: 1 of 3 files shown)


SaintShit:
Hi, thanks for your participation. This feature lets everyone know data of any user, right?
Actually this bot were supposed to be an admin bot only, but this kinda feature would be useful too. how about the security so?

bahram1249:
In fact, the data that we show to general users is not very important for others. Therefore, it can be ignored because observing the bandwidth consumption does not disturb the security of users. Although this data is not completely public and we have to send the intended username.

PR #55, opened 2023-02-28
Fix various issues

alsuo:
I tried the `hosts-front` branch, but the dashboard was not working correctly and had a few issues. This PR fixes those issues.

- Backend was sending `{"users": [{...}, {...}, ...], "total": ...}` in `/api/users` but frontend was expecting the old structure without count (`[{...}, {...}, ...]`).
- There was a useless API call to `/api/users` to fetch users in login screen and after logging in, there was no API call to fetch users. So, the dashboard displayed no user after login, and a page refresh was needed to see users.


alsuo:
Also I have a question.

Is there any specific reason to remove the "Reset Every" feature from front-end in this branch?

SaintShit:
Hi, nice job. the API change was recently made and @gozarman was going to work on it. he's gonna complete the pagination part.


> Also I have a question.
> 
> Is there any specific reason to remove the "Reset Every" feature from front-end in this branch?

Yes. @gozarman decided to re-implement it.

PR #46, opened 2023-02-28
Custom sni and host via XRAY_HOSTS

alsuo:
With this PR, we can set a custom `sni` or `host` for each host in `XRAY_HOSTS`. I have tested it and works fine both in dev and in production.

Two syntaxes are supported for `XRAY_HOSTS` lines:

1. The simple `remark@address` syntax.

2. JSON syntax:
   ```
   {"remark": "myremark", "address": "10.20.30.40", "sni": "mysni.example.com", "host": "myhost.example.com"}
   ```

Example `.env`:

```
XRAY_HOSTS = '
Server@127.0.0.1
Bridge@192.168.1.10

# Empty lines and comments are supported
John\'s Domain@john.example.com

# You can set custom sni and host using JSON format:
{"remark": "CDN2", "address": "10.20.30.40", "sni": "my.example2.com", "host": "my.example2.com"}
'
```

PR #57, opened 2023-02-28
Add "sni" and "host" fields to hosts

alsuo:
Add `sni` and `host` fields to `hosts` table and hosts dialog. See #46


SaintShit:
I sincerely appreciate your efforts. <3

PR #70, opened 2023-03-05
Easily Add Expire Date

bahram1249:
now with this pieces of code you can add expire date in telegram robot using regex for (x) months or years later.


bahram1249:
Examples : 
2m :  set expire date for 2 months later from now
1m :  set expire date for 1 years later from now

PR #103, opened 2023-03-18
Update system.py

reloadlife:
d was missing in bandwidth


SaintShit:
Hi, thanks. please apply fix to all places including SystsemStats model and front end

SaintShit:
any update on this?

PR #104, opened 2023-03-18
 Marzban CLI: import-from-env implemented

Alirezaja1384:
Details:
  - It imports env's sudo admin to the database
  - It sets admin_id for previously created users which their admin_id is NULL


SaintShit:
This is awesome, thanks.

PR #110, opened 2023-03-19
Migration conflict resolved

Alirezaja1384:
case_insensitive_username and admin_is_sudo_field conflict resolved

Details:
  - fad8b1997c3a's down revision set to 0d3d400ea75

### DIFF
--- a/app/db/migrations/versions/fad8b1997c3a_case_insensitive_username.py
+++ b/app/db/migrations/versions/fad8b1997c3a_case_insensitive_username.py
@@ -11,7 +11,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = 'fad8b1997c3a'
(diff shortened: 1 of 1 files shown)


SaintShit:
Yo I just got that you've changed the down revision manually and I mistakenly merged it. better approach of merging two heads is to run `alembic merge heads` command. this generates a new migration scripts autmatically and fixes the conflict.
changing the down revision manually might cause your migration script to be ignored if someone has run the migration scripts of dev version before.

Alirezaja1384:
That's right. Sorry about the inappropriate fix, and thank you for the information. As I said, I don't have that much experience with sqlalchemy. I prefer Django ORM myself. It's way easier for me :)
Also, I think you could revert the pull request. It might be a better option because others will know the merge was by mistake.

PR #127, opened 2023-03-25
BugFix: keep count of added proxies in `added_proxies` before committing the changes

govfvck:
the bug was that when it tried to exclude inbounds, it looked up the database and checked if any proxy needed inbound exclusion. the problem was when a proxy was just added and wasn't committed to the database yet, so it was thought the proxy didn't exist.


SaintShit:
exactly right, thanks for your notice.

PR #147, opened 2023-03-29
ClashConfiguration's add_vmess: network bug fixed

Alirezaja1384:
Details:
 - "network" option should be specified regardless of tls.


SaintShit:
Thanks.

PR #160, opened 2023-04-03
Clash template implemented

Alirezaja1384:
Clash template and its utilities implemented.

Details:
  - Jinja2 installed
  - app/templates: render_to_string() implemented
  - app/templates/filters.py: `yaml`, `except`, and `only` filters implemented
  - `CUSTOM_TEMPLATES_DIRECTORY` and `CLASH_SUBSCRIPTION_TEMPLATE` configs added
  - ClashConfiguration: Template rendering implemented
  - share.py: generate_subscription() implemented
  - Default template `clash/default.yml` added
  - Subscription CLI app added
  - CLI subscription: get-link and get-config commands implemented
  - cli/utils.py: paginate() and get_user() implemented
  - CLI documentation updated

Side notes:
  - More custom configuration files added to .gitignore
  - `PASSWORD_ENVIRON_NAME` and `FLAGS` moved to cli/utils.py


Alirezaja1384:
I thinks it's finished now. If there's any bug or something needs to be improved, I'll be happy to do it.

SaintShit:
Hi, I checked the changes. I think this is great. I really appreciate your job. however I'm not still sure about having a templates directory just for Clash configuration.
I was thinking that other options such as `mode'`, `port` and `proxies` should also be there. what your idea?

SaintShit:
Please add the jinja2 to the requirements too.

Alirezaja1384:
Thank you for your review.
I don't know if there's a better option for the templates directory. Jinja can use also strings as a template so it can be stored in the database or whatever else, but I don't know if they're reliable enough for upcoming features.
Personally, I just use `proxies` options and handle the rest myself (like [this](https://gist.github.com/Alirezaja1384/03d1670c33e0feaa2783196ff626d529)). I also think it would be better to move `mode` and `port` to `default.yml` so the user won't accidentally use the default options. Just like how I did for the `♻️ Automatic` proxy group.

Alirezaja1384:
> Please add the jinja2 to the requirements too.

I think I did it. As I see [here](https://github.com/Gozargah/Marzban/pull/160/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R64).

SaintShit:
> > Please add the jinja2 to the requirements too.
> 
> I think I did it. As I see [here](https://github.com/Gozargah/Marzban/pull/160/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R64).

yay the problems was my environment

Alirezaja1384:
I think it makes more sense now. IMO context objects should only contain dynamic options like `proxies`, `proxy-groups`, and `rules` that might be generated according to the user's configurations.

Alirezaja1384:
I'm still thinking about templates. There might be some use cases, for example, if we want to allow users to customize certain CLI outputs, they can be really useful. For example, if the user wants to extract users' data as json/html/markdown and use it for some sort of scripts/representation or anything else. Its use cases might be rare, although can make some parts of the app much more customizable. Personally, I don't know if there's a better way of storing templates or not (at least yet), but I look forward to it.

SaintShit:
Hi, your changes are great. I'll definitely merge it in the coming days, just wanted to inform you. Sorry for being late.

PR #177, opened 2023-04-09
fix: dashboard route added to .env file

Navid079:


SaintShit:
Hi, this won't work in that way. front-end is built for the /dashboard path, so that to change the dashboard path, you need to build the front-end again. changes must be applied to front-end first.

PR #196, opened 2023-04-20
Backend: Improve server statistics

### LINKED ISSUE #134: Server statistics improvement
- [ ] CPU usage
- [ ] CPU cores
- [ ] Real-time traffic transfer status

govfvck:
fixes #134 

add `cpu_cores` and `cpu_usage` to get number of cpu cores and percentage of cpu usage.

add `incoming_bandwidth_speed` and `outgoing_bandwidth_speed` to show real time data about system's bandwidth
	add utility functions to monitor outgoing and incoming bytes/packets
	realtime data is available in `rt_bw` inside the `utils/system.py`,  and it will be updated every second

### DIFF
--- a/app/models/system.py
+++ b/app/models/system.py
@@ -4,7 +4,11 @@ from pydantic import BaseModel
 class SystemStats(BaseModel):
     mem_total: int
     mem_used: int
+    cpu_cores: int
+    cpu_usage: float
     total_user: int
     users_active: int
     incoming_bandwidth: int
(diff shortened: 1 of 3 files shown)


Musixal:
Nice work. But without frontend implentation it's not functional right now.

govfvck:
yes, i did some tests and it was looking fine as far as my tests go.
you're right, but this needed to be done on the backend side so that maybe someone will take care of it on the front side :)

PR #207, opened 2023-04-23
support change detail(password, id, etc.) of protocol

zhongfq:
This commit is used to implement the change protocol details(eg. password, id, method, flow, etc.).

Because this is my first time to use ts and react, I'm not familiar with the 'Select' control. Use 'Select' to implement 'method of Shadowsocks' is better. 

IMG_8AD6944A5F99-1


SaintShit:
Man this is great thank you!! I think I need to do a validation on backend to check the uniqueness of the passwords, I'm going to do this now. since the changes are on the frontend, I want @gozarman to check it.

PR #230, opened 2023-04-30
docs: marzban-node usage

zhongfq:


SaintShit:
I'm really thankful for the favor you did. I want you to make some changes if you can please
this section would be nice if it is in this format:
‍‍‍‍‍‍‍‍```
- install with docker
- manual install 
| - using systemctl
```

Also, marzban-node (starting with lower-case) should be cloned in `/opt` directory

zhongfq:
why marzban-node placed in '/opt', but marzban placed '/var/lib'?

SaintShit:
`/var/lib` directory is used to store application data. they'll remain the same if marzban's setup or configuration is changed. while `/opt` directory is supposed to contain application files and add-ons.
also we've prepared a new installation script called [marzban.sh](https://github.com/Gozargah/Marzban/tree/dev#marzbansh), it downloads marzban setup files ( compose, json config and etc. ) in `/opt` directory as well.

indeed, marzban-node should be downloaded in `/opt/marzban-node` and it's data files ( currently ssl_cert.pem and ssl_key.pem ) must be in `/var/lib/marzban-node`

PR #246, opened 2023-05-05
2 Minor Changes

mahanmi:
This is a very useless PR I have just changed a button name and updated the xray installation version in the readme files


SaintShit:
Hi. thank you, but you shouldn't make your front-end changes on the build files, you have to edit the components and re-build.
Also, we still recommend installing Xray using Xray-install, our script just fetch the binary.

mahanmi:
I'm Sorry for my mistakes I'm new here I'll try to do that

PR #247, opened 2023-05-06
config 'VITE_BASE_API' by user

zhongfq:
This make it can  debug on iPad, iPad or other devices.


SaintShit:
Yes! you're doing nice. thanks

PR #249, opened 2023-05-07
i18n: hosts setting

zhongfq:


SaintShit:
Thanks. I'm thinking about changing i18n file to separate files and something ez to edit by everyone. plz tell if u have any idea on this

zhongfq:
> Thanks. I'm thinking about changing i18n file to separate files and something ez to edit by everyone. plz tell if u have any idea on this

it is good idea. if can support this format, may be better:

login: {
   en: "Login",
   zh: "登录",
}

SaintShit:
> > Thanks. I'm thinking about changing i18n file to separate files and something ez to edit by everyone. plz tell if u have any idea on this
> 
> it is good idea. if can support this format, may be better:
> 
> login: { en: "Login", zh: "登录", }

I checked i18next docs, it has multiple backends support. We're searching for a gui tool for these jsons, that'll be more comfortable for everyone to do the translation then.

PR #295, opened 2023-05-30
Persian language translation

MuhammadAshouri:


M03ED:
i tried to update marzban but its not gonna be update 
i want to test persian
image
i want to test persian

PR #326, opened 2023-06-07
Update Subscription index.html

M03ED:
DateTime was based on Coordinated Universal Time, but this was not mentioned


M03ED:
image

SaintShit:
Hi, the image of apps are not loaded, and I think it'd be better to load them from their repository or their app store image

M03ED:
i think about this but i didnt remember any service we can upload photos to use them 
google play is blocked in iran right now and there is chance app store be blocked again so we cant use them
so i choose github to do this
also i add clash verge and add both googleplay and github link for users dont have access to free internet

M03ED:
image

M03ED:
i think we don't need this pr anymore.

PR #350, opened 2023-06-18
bot database backup option

NotMmDG:
added database backup for admin side of telegram bot

### DIFF
--- a/app/telegram/handlers/admin.py
+++ b/app/telegram/handlers/admin.py
@@ -2,7 +2,8 @@ import io
 import math
 import re
 from datetime import datetime
-
+import os
+import time 
 import qrcode
 import sqlalchemy
 from dateutil.relativedelta import relativedelta
@@ -131,7 +132,15 @@ def suspend_user_command(call: types.CallbackQuery):
         reply_markup=BotKeyboard.confirm_action(
             action="suspend", username=username),
     )
-
+@bot.callback_query_handler(cb_query_equals("backup"), is_admin=True)
+def backup(call: types.CallbackQuery):
+    bot.edit_message_text(
+        'Are you sure? this will send you the backup once every 4 hours.',
+        call.message.chat.id,
+        call.message.message_id,
+        reply_markup=BotKeyboard.confirm_action(action="back_up")
+    )
+    
 
 @bot.callback_query_handler(cb_query_startswith("activate:"), is_admin=True)
 def activate_user_command(call: types.CallbackQuery):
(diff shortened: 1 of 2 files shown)


M03ED:
few changes can make it better

Make time change able.

read database path from .env file

use different bot token ( it can be better )

i write code for it you can use it if you want

import os
import shutil
import time
import datetime
import requests
import threading

# Set up Telegram bot token and chat ID
bot_token = 'AAAAAAAAAA:11111111111111111111111111'
chat_id = '1111111111'
sec = 7200

def backup_and_upload():
    while True:
        # Get the current timestamp
        timestamp = datetime.datetime.now().strftime("%Y-%m-%d-%H:%M")

        # Create the source and destination file paths
        source_file_path = '/var/lib/marzban/db.sqlite3'  # Replace with the actual source file path
        destination_file_path = f'{timestamp}.sqlite3'  # Replace with your desired file extension

        # Delete the previous file if it exists
        if os.path.exists(destination_file_path):
            os.remove(destination_file_path)

        # Copy the source file to the destination path with the new name
        shutil.copy2(source_file_path, destination_file_path)

        # Open the copied file in binary mode
        with open(destination_file_path, 'rb') as file:
            # Prepare the Telegram bot API endpoint
            url = f'https://api.telegram.org/bot{bot_token}/sendDocument'

            # Create the request payload
            files = {'document': file}
            params = {'chat_id': chat_id}

            # Send the file to the Telegram bot using the SOCKS proxy
            response = requests.post(url, params=params, files=files) 

        # Print confirmation message
        if response.status_code == 200:
            print(f"Uploaded {destination_file_path} to Telegram bot.")
        else:
            print(f"Failed to upload {destination_file_path} to Telegram bot.")

        # Wait for 1 hour before repeating the process
        time.sleep(sec)

# Create a new thread for the backup and upload task
backup_thread = threading.Thread(target=backup_and_upload)

# Start the thread in the background
backup_thread.daemon = True
backup_thread.start()

# Keep the main thread running
while True:
    pass

MuhammadAshouri:
@M03ED Why did't you read bot token and chat id from .env

NotMmDG:
so you need another bot token and admin parameter?
wouldn't it better to use the same bot to send the details into a group maybe?

SaintShit:
Hi, thanks for contribution. I must say that what you did is not enough. some might use MySQL or something else, we need a method or command that bumps database for backup. I think bumping database to CSV or SQL file would be nice. btw, I'm going to close this PR and add it as a task to board.

PR #385, opened 2023-07-13
Reseting user's proxy ids(uuid/password) implemented

Alirezaja1384:
After revoking the user's subscription the old one should be useless, without changing the proxy ids, the user can still connect without any problem and I don't think it's "revoking the subscription". Anyway, I might be wrong, if I am, feel free to close the PR 😄.

Details:
 - crud: revoke_sub() resets proxy ids by default.
 - `/api/user/{username}/revoke_sub`: `reset_ids` query parameter added.

### DIFF
--- a/app/db/crud.py
+++ b/app/db/crud.py
@@ -287,8 +287,23 @@ def reset_user_data_usage(db: Session, dbuser: User):
     return dbuser
 
 
-def revoke_user_sub(db: Session, dbuser: User):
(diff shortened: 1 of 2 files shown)


SaintShit:
@Alirezaja1384 Hi, no u're not wrong. I asked about it in group and had planned to implement this, which you have done generously, thank you <3 imma merge this as soon as I got my PC available

PR #401, opened 2023-07-23
Fixing node.py

M03ED:
there was something wrong in last pr so i close it.
i change file so not sudo admin couldn't see nodes detail
but there is a bug in webpage
if not sudo admin go in node settings it will be logout after few second
im not good at frontend and i don't know how to fix it and i think its better @gozarman fix it

### DIFF
--- a/app/views/node.py
+++ b/app/views/node.py
@@ -53,6 +53,8 @@ def add_node(new_node: NodeCreate,
 def get_node(node_id: int,
              db: Session = Depends(get_db),
              admin: Admin = Depends(Admin.get_current)):
+    if not admin.is_sudo:
+        raise HTTPException(status_code=403, detail="You're not allowed") 
 
     dbnode = crud.get_node_by_id(db, node_id)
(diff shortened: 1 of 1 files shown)


M03ED:
i found a way to solve frontend problem with node.py file.

SaintShit (commented):
this exception must be caught at front-end, so this change is not valid

app/views/node.py
```diff
@@ -131,7 +131,7 @@ async def node_logs(node_id: int, websocket: WebSocket, db: Session = Depends(ge
 def get_nodes(db: Session = Depends(get_db),
               admin: Admin = Depends(Admin.get_current)):
     if not admin.is_sudo:
-        raise HTTPException(status_code=403, detail="You're not allowed")
```
SaintShit: this part I mean

SaintShit:
so I'm going to cherry-pick 56e711b53173a48661da6d3a0157ad98a953ecf4 only

PR #423, opened 2023-08-01
Marzban Cli Improvment

M03ED:
i add deleting expired users , increase and decrease data limit and expire time for all users at same time

### DIFF
--- a/app/db/__init__.py
+++ b/app/db/__init__.py
@@ -26,7 +26,7 @@ def get_db():  # Dependency
 from .crud import (create_admin, create_user, get_admin, get_admins,  # noqa
                    get_jwt_secret_key, get_or_create_inbound, get_system_usage,
                    get_user, get_user_by_id, get_users, get_users_count,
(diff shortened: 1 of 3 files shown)


M03ED:
i moved is_user_expired to crud.py to be access able with api , i will add these functions on api soon.

SaintShit:
Hello, thank you for working on Marzban. the problem is that this approach cannot do xray operations, as xray is runned in another process. it just can apply changes to database user and Marzban must be restarted for the changes to take effect. so, calling xray.operations is wrong.

M03ED:
a question 
we just change expired user and they are disabled already , if we don't call xray at all gonna make a bug in xray ?

SaintShit:
> a question we just change expired user and they are disabled already , if we don't call xray at all gonna make a bug in xray ?

not, it doesn't need to call xray operations in this case.

M03ED:
ok i fix that and add these functions to api , pr is complete and ready to merge

PR #520, opened 2023-09-05
Custom update interval and new variables for subscription

kastov:
Add following headers:
- Support-Url
- [Profile-Web-Page-Url ](https://docs.cfw.lbyczf.com/contents/urlscheme.html#profile-web-page-url)
- Profile-Title

New .env variables SUB_PROFILE_TITLE, SUB_SUPPORT_URL, SUB_UPDATE_INTERVAL

So you can set your own Profile-Title, Support-Url and Update Interval (default is 12 hours)


app/utils/share.py
M03ED, line 1: i think you need to update this file , in last commits this file changed a lot

kastov: Okay. No problem. I will close this PR and create a new one, based on latest commits

kastov: Not sure that i merged correct. But seems like now conflict resolved.

septs:
is there any documentation on the `profile-title` and `support-url` header fields?

PR #546, opened 2023-09-15
fix geoip and geosite download problem

oXIIIo:
- fix the issue of not being able to download geoip and geosite files in the sing-box template
- format the template files with the sing-box formatter

### DIFF
--- a/app/templates/singbox/default.json
+++ b/app/templates/singbox/default.json
@@ -1,10 +1,8 @@
 {
   "log": {
-    "disabled": false,
     "level": "warn",
     "timestamp": true
   },
-  "experimental": {},
   "dns": {
     "servers": [
       {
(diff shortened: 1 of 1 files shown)


M03ED:
This is a good merge, but there are other bugs in the 'sing box' that need fixing. We are waiting to identify all the bugs before fixing them all at once. Also, you removed some unused values from the template. I added them to the default template for people who want to personalize it and see their options.

M03ED:
And for geoip file problem we just need to add this line 
      "download_detour": "direct"

oXIIIo:
As I mentioned, it has been formatted by the "sing-box" formatter, and for this reason, some unnecessary initializations have been removed. Additionally, in the end, users need to read the "sing-box" document to understand the purpose of each section.
as you can see clash config is minimal as possible.

master88wi:
sing-box good

PR #556, opened 2023-09-20
Refining xtls-rprx-vision Flow Selection in TCP Configurations

thenorthatlantic:
If we select "xtls-rprx-vision" as the flow, it was [previously ](https://github.com/Gozargah/Marzban/issues/478)enabled for all TCP configurations in vless, regardless of whether they were using "reality" or not. However, I have made modifications to the code to ensure that it is now exclusively applied to Reality + TCP configurations, excluding other types such as TCP + HTTP or TCP + TLS.

### DIFF
--- a/app/utils/share.py
+++ b/app/utils/share.py
@@ -104,8 +104,8 @@ class V2rayShareLink(str):
             "type": net,
             "host": host,
             "headerType": type
-        }
-        if flow and net in ('tcp', 'kcp'):
+        }             
+        if flow and net == 'tcp' and tls == 'reality':
             payload['flow'] = flow
 
         if net == 'grpc':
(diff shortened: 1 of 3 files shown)


M03ED:
Your pr is not good and it need some changes , xtls it's not only for reality and also should be applied for kpc 
I ask you xray configuration in #478 but you didn't answered.

thenorthatlantic:
> Your pr is not good and it need some changes , xtls it's not only for reality and also should be applied for kpc I ask you xray configuration in #478 but you didn't answered.

I think xray has not featured xtls support for KCP since version 1.8.0.

M03ED:
This configuration problem it's not gonna be solved with your change and needs more change , I will try to fix it in next week

thenorthatlantic:
> This configuration problem it's not gonna be solved with your change and needs more change , I will try to fix it in next week

Sorry I wasn't entirely proficient in the panel's code. I appreciate your help Thank you.

PR #621, opened 2023-11-05
public name for node , get node usage with token

M03ED:
now user can see his usage with using `/sub/{token}/usage`


Iambabyninja:
It's good 👍

PR #625, opened 2023-11-08
Increase username limit to 64 characters

lilcheti:


app/db/models.py
M03ED, line 1: it must be 64 not 66

M03ED (changes requested):
database changes need migration file to sync database with code and you need to make and upload it with code 
also you did this change on master branch , changes and new features first come on dev branch then come to master branch

kastov:
Does xray supports that lenth?

app/db/models.py
lilcheti: It was 34 how is that?

kastov:
> > Does Xray-core supports that lenght?
> 
> A valid mail's username length is 64 characters 64 (Username) + 1 (@) + 255 (Domain Name) = 320 [i.e. example@localhost ]
> 
> How can Xray-core not accept that?

Not sure about 320 in total. Many sources say that the maximum length is around 254-256 characters.

SaintShit:
this isn't enough, you need to provide a migration script too

PR #630, opened 2023-11-10
Install xray-core from local script

APT-ZERO:


M03ED (commented):
Not useful. 
Xray script do all these automatically, and there is no need to do it manually.

PR #638, opened 2023-11-17
User Online/Offline Status in Panel

x0sina:
1. User Last seen added in table
2. If no data it shows no data
3. If user is connected in less than 1 minutes user status is online


M03ED:
i think you forgot to upload build folder , you should build that and upload it

PR #640, opened 2023-11-17
Add User Status online/offline - Last Seen

x0sina:
photo_2023-11-17_20-36-51

1. If User online_at data is less than 60 seconds to now user is online
2. If User online_at data is more than 60 seconds to now user is offline - and can access last seen
3. If User online_at data is null we have nodata option


SaintShit:
Nice, thank you!

VladimirFedulov:
It doesn't work for me with VLESS TCP REALITY

x0sina:
This Pull Request is only front-end 

> It doesn't work for me with VLESS TCP REALITY

This commits is only client side and type of config is not important because data of user online status will sent from backend and database, it should work with any type of configs

x0sina:
> Nice, thank you!

Thanks for merging❤️

kastov:
> > Nice, thank you!
> 
> Thanks for merging❤️

#644 can you check please? Seems like your PR did not work correctly for some users.

PR #692, opened 2023-12-11
Add / modify user by template

ImMohammad20000:
now we can add or modify user by **templates** using `POST /api/user/from_template` and `PUT /api/user/{username}`

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -96,6 +96,19 @@ class User(BaseModel):
         return v
 
 
+class CreateUserFromTemplate(BaseModel):
+    username: str
+    template_id: int
+
+    @validator("username", check_fields=False)
+    def validate_username(cls, v):
+        if not USERNAME_REGEXP.match(v):
+            raise ValueError(
+                "Username only can be 3 to 32 characters and contain a-z, 0-9, and underscores in between."
+            )
+        return v
+
+
 class UserCreate(User):
     username: str
     status: UserStatusCreate = None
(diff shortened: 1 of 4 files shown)


app/utils/share.py
M03ED, line 1: share.py its a file for making configs for users not creating new user

app/models/user.py
M03ED, line 1: you can add property here and make user in here at CreateUserFromTemplate class

app/views/user.py
```diff
@@ -6,13 +6,20 @@
...
+    UsersResponse,
+    UserStatus,
+    UserUsagesResponse
+)
+from app.utils import report, share
 
 
 @app.post("/api/user", tags=['User'], response_model=UserResponse)
-def add_user(new_user: UserCreate,
+def add_user(new_user: UserCreate | CreateUserFromTemplate,
```
M03ED, line 22: its better to separate api requests

requirements.txt
```diff
@@ -44,6 +44,7 @@ PyMySQL==1.0.3
 pyOpenSSL==23.2.0
 pyparsing==3.0.9
 pypng==0.20220715.0
+PySocks==1.7.1
```
M03ED, line 47: i can't find where you use this new library

M03ED (commented):
i don't think this be approved because templates are useful when they come to dashboard, changes like this is good for when we can set different access for each admin

requirements.txt
ImMohammad20000: i try to access telegram use socks and that needed this lib so i added

app/utils/share.py
ImMohammad20000: don't want to make routers too long

PR #704, opened 2023-12-19
converted xray functions to async, to make logs api sane

parhawm:
I had some problems with Marzban logs API, so I re-wrote some parts asynchronously using anyio for sharing the logs.
This fixes some issues with reading logs from ws, looked like the data was corrupted sometimes.
however I think it needs some front-end work.

### DIFF
--- a/app/jobs/0_start_xray.py
+++ b/app/jobs/0_start_xray.py
@@ -1,25 +1,26 @@
 import traceback
+import asyncio
 
 from app import app, logger, scheduler, xray
-from app.utils.concurrency import threaded_function
 
 
-def core_health_check():
-    if not xray.core.started:
-        xray.core.restart(xray.config.include_db_users())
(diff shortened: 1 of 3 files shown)


SaintShit:
there are some problems in logs ws, but I think it has more to do than making functions async, as the whole application isn't written async.

parhawm:
I've actually rewrote everything except database in async including nodes; checkout my [grpclib branch](https://github.com/parhawm/Marzban/tree/grpclib)
there could be a discussion about using async or threading

However about this specific pr, I still haven't understood where the problem is with sharing logs using a deque between two threads in this way, so I can't fix the current code
but I've tested using two websocat connections, by using diff to ensure they're the same, they were NOT.
This fixes that problem, using anyio thread safe streams, I've tested this and it works 100% correctly.

PR #724, opened 2023-12-30
Some Changes and fixes

M03ED:
- [Restrice None Sudo Admin access](https://github.com/Gozargah/Marzban/commit/4ba839817bac4b49e0071af1614ffc04d21bb77e)
- [Fix sending duplicate notfication](https://github.com/Gozargah/Marzban/commit/0f01b5d501df9cda91234ca17aca854e5bc19927)
- [add on_hold icon to language files](https://github.com/Gozargah/Marzban/commit/a4b8ad92a16ee9e20c5a45251e220b1eac53f41a)
- [Send random sub domain for subscription url](https://github.com/Gozargah/Marzban/commit/accd7da9a88a85d3833df955659af460d824ba5c)
- [[Fix] Emoji Status](https://github.com/Gozargah/Marzban/commit/548828f459a10930d7f2e438dfe4a854a8f95ca3)
- [[Fix] Header in clash and sing-box](https://github.com/Gozargah/Marzban/commit/bf58b1c374f0234efd21f3f5698087d50b354cf7)
- Make * subdomain more random


app/jobs/review_users.py
```diff
@@ -34,7 +35,7 @@ def add_notification_reminders(db: Session, user: "User", now: datetime = dateti
                 user.id, user.expire)
 
 
-def review():
+def review(bg: BackgroundTasks):
```
SaintShit, line 38: Passing a bg instance isn't a good idea. as mention [here](https://www.starlette.io/background/), if one of the tasks raises an exception, the next tasks will not execute. so I would rather use threading for this, or at least have an instance of BackgroundTasks in each run. so please change this section.

app/jobs/review_users.py
```diff
@@ -77,10 +78,23 @@ def review():
...
+            active = user.expire and datetime.timestamp(user.expire) >= now.timestamp()
+            if active:
+                status = UserStatus.active
+            else:
+                continue
+
+            update_user_status(db, user, status)
+            xray.operations.add_user(user)
+            
+            logger.info(f"User \"{user.username}\" status fixed.")
```
SaintShit, line 97: I previously removed this section because I think it must be fixed where it's happening rather than writing an extra check. It rarely happens that a user gets expired by mistake ( never happened to me ), I appreciate any evidence or clue to finding the problem.

PR #789, opened 2024-02-09
feat(node): Get Node stats

M03ED:
we can get nodes stats with api for better managment


M03ED:
marzban side for [this](https://github.com/Gozargah/Marzban-node/pull/34) pr

PR #795, opened 2024-02-15
feat(cli): added admin usage to table of admins

danielsoheil:


danielsoheil:
image

danielsoheil:
With this database structure it's a little bit complicated, I think it's not worth it.

cli/admin.py
```diff
@@ -34,6 +35,16 @@ def validate_discord_webhook(value: str) -> Union[str, None]:
     return value
 
 
+def calculate_admin_usage(admin_id: int) -> int:
+    with GetDB() as db:
+        return db.execute(text(f'select sum(used_traffic) / 1024 / 1024 / 1024 from (select * from node_user_usages join (select * from users where admin_id = {admin_id}) as users on users.id = node_user_usages.user_id);')).one()[0]
```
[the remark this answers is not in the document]
M03ED: also you can disable node_user_usages with env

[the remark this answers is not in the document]
danielsoheil: > why you use node_user_usages table for calculating ? we already have used_traffic column in users table

why we have same data in two different location? which one is SSOT?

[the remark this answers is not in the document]
danielsoheil: > also you can disable node_user_usages with env

I can not understand, can you explain more?

danielsoheil:
> Hi, that's great, could you please use sqlalchemy's API rather than executing raw query?

done.

cli/admin.py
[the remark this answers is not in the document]
danielsoheil: > > > why you use node_user_usages table for calculating ? we already have used_traffic column in users table
> > 
> > 
> > why we have same data in two different location? which one is SSOT?
> 
> node_user_usages it's same as used_traffic but it's separated for each node Also if you remove a node that's gonna be deleted too But used_traffic will be changed if you reset usage

Okey
I used used_traffic in users table.

app/db/models.py
M03ED, line 1: why we have duplicate function ? there both do same thing also we have lifetime_used_traffic function and you can calculate reseted_usage way faster
`reseted_usage = lifetime_used_traffic - used_traffic`

cli/admin.py
M03ED, line 1: use power instead of Multiplication
`usage // (1024 ** 3)`

M03ED (changes requested):

cli/admin.py
M03ED: actually its better to use `from app.utils.system import readable_size` function

app/db/models.py
danielsoheil: that's not same, you have a better idea?

lifetime_used_traffic is not faster :/
lifetime_used_traffic will get all data from database and calculate it in python
by reseted_usage that I add, you will get calculated data from database

M03ED: we are doing same thing twice, you can do less job
Another way to avoid more jobs is that we calculate rested usage then for  lifetime_used_traffic sum rested usage with used traffic to calculate lifetime_used_traffic

danielsoheil: you can do it yourself.

danielsoheil:
I implemented this feature for my self but one of my friends need this. i told her this is going add in official version, when do you think you can work on it?

danielsoheil:
guys, i need this
implement it your self or merge this

app/db/models.py
tohid3aeed: من در مورد نحوه محاسبه ریست یوزیج ابن کار رو به صورت دستی انجام میدم
مصرف کل به علاوه مانده حجم که مقدار کامل رو درمیارم
برای مثال 
اکانت ۵۰ گیگ دوبار کامل مصرف شده ریست شده و حجم کل ۱۳۰ گیگ هستش
۱۳۰ گیگ به اضافه ۲۰ گیگ مانده از ۵۰ گیگ 
عکسی که پیوست کردم رو من حجم رو 1100  حساب میکنم
IMG_20240315_002806_454

realartin:
> guys, i need this implement it your self or merge this

ما هم همینطور بهش نیاز داریم

SaintShit:
it's merged to `dev`

PR #806, opened 2024-02-20
Feat(Subscription): v2ray Custom config

M03ED:
- add v2ray Custom config
- move all mux templates to single file
- move different config generating class to multi file
- update host settings in front by @MuhammadAshouri 
- add fragment support for v2ray config


M03ED:
> And standart does not include([XTLS/Xray-core#716 (comment)](https://github.com/XTLS/Xray-core/discussions/716#discussioncomment-6272921) `fragment` part.

we choose streisand standard for this `&fragment=length,interval,packet`
if you have connection with foxray developer ask him to make it like this

llllnima1011:
Tnx Alot :D

M03ED:
> @M03ED
> 
> My dear friend,
> 
> For example, here is a link, which, when added in a single JSON, can contain multiple values.
> 
> streisand://import/https://marzban.dev/test.json#testSub
> 
> Here is the file itself separately
> 
> https://marzban.dev/test.json
> 
> However, there is a REMARK parameter that is only supported in v2rayNG so far. The parameter sets the name of this connection in the list. Pls, add its 2 your commit. Streisand say, what: image

i already add remarks field in this pr

PR #819, opened 2024-02-25
fix(timezone): remove using deprecated utcnow.

Jolymmiles:
May fix error with subscription time and logs.

Now it will use srver timzone insted hardcoded utc.


M03ED:
using server time zone its not correct and may cause Lack of coordination between front and backend , if you wanna do this you need way more changes in front and backend.
also limiting memory for program like marzban that may need lots of memory its not correct.
i close this pr but if you can fixt first issue to not make Lack of coordination that may be accepted.

PR #842, opened 2024-03-03
fix: tcp http in v2ray json config

M03ED:


ImMohammad20000:
please implement random User-Agent function for User-Agent field

fodhelper:
I have tested this PR and problem fixed
thank you

PR #882, opened 2024-03-20
Support httpUpgrade

M03ED:
- add httpupgrade support
- mux can be disabled depend on config type ( in some case mux may cause conflict in sing-box ) @fodhelper
- add mux for clash

### DIFF
--- a/app/subscription/clash.py
+++ b/app/subscription/clash.py
@@ -1,7 +1,8 @@
 import yaml
+import json
 from app.templates import render_template
 
-from config import CLASH_SUBSCRIPTION_TEMPLATE
+from config import CLASH_SUBSCRIPTION_TEMPLATE, MUX_TEMPLATE
 
 
 class ClashConfiguration(object):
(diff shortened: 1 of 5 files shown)


iambabyninja:
Big thanks, it's cool 🤝❤️

fodhelper:
clash does not support any mux, it's clash-meta that supports mux
Isn't it better to change it's name to clash-meta like it's subscription name?
https://github.com/Gozargah/Marzban/blob/master/app/views/subscription.py#L193C64-L193C74
@M03ED

M03ED:
> clash does not support any mux, it's clash-meta that supports mux
> Isn't it better to change it's name to clash-meta like it's subscription name?
> https://github.com/Gozargah/Marzban/blob/master/app/views/subscription.py#L193C64-L193C74
> @M03ED 

Official clash client is dead , docs are deleted and it's not gonna be supported anymore so it don't have much different

PR #890, opened 2024-03-25
Optional json sub

thenorthatlantic:
Added a simple 'env' option to allow for optional JSON configuration in subscriptions.


M03ED:
may i ask why you avoid using custom config ?
also you can do same thing with adding `/v2ray` to end of your sub link

thenorthatlantic:
@M03ED it's advantageous to have that option as optional.
In v2rayNG, custom configurations lacks the QR code share feature.
When using the /v2ray sub, users encounter difficulty accessing the template page.

I also had an idea of checking if V2RAY_SUBSCRIPTION_TEMPLATE was not set to "false" and providing JSON, but I wasn't sure if that would conflict with other parts of the code. If you believe that checking V2RAY_SUBSCRIPTION_TEMPLATE wouldn't cause problems, I will modify the code accordingly, so there is no need to add extra option in env.

M03ED:
> @M03ED it's advantageous to have that option as optional.

This can't be called an advantage
Costum configs have way more advantages 
Using fragment and mux for single config instead of all configs, allow insecure , make ir domains direct and lot's of more options you can have
You can use custom config and set these settings for your users easily or you need to have physical access to there mobile to do them manually 

> In v2rayNG, custom configurations lacks the QR code share feature.

Share sub link instead of config

mikeesierrah:
You are right custom configs are way better and I prefer them 

But I understand his pov and he does have point 
I hope marzban doesn't lose flexibility
by forcing a feature that some people might not want to have , for any reason,
And it's not like it will mess with GUI , it's a part of env
the code is ready too so it doesn't need you guys to write it from scratch, it's just has to be merged

mikeesierrah:
@M03ED You are right custom configs are way better and I prefer them

But I understand his pov and he does have point
I hope marzban doesn't lose flexibility
by forcing a feature that some people might not want to have , for any reason,
And it's not like it will mess with GUI , it's a part of env

M03ED:
> @M03ED You are right custom configs are way better and I prefer them
> 
> But I understand his pov and he does have point I hope marzban doesn't lose flexibility by forcing a feature that some people might not want to have , for any reason, And it's not like it will mess with GUI , it's a part of env

we discuss about this before , check #782

M03ED:
check this one too 
https://github.com/XTLS/Xray-core/pull/3076#issuecomment-2017409517

fodhelper:
This can be useful for v2rayN, because custom json will disable it's url-test/real-delay
but for v2rayNG, i see no important reason

SaintShit:
Personally, I prefer custom json config to be disabled by default. I like this configuration variable, but the implementation is a little bit wrong.
Please rename the variable to "USE_CUSTOM_JSON_DEFAULT" and set it's default value to False. Also, add some detail on the variable comments that what this supposed to do.

SaintShit:
custom json is also supported in Streisand, we should add support for it too.

thenorthatlantic:
@SaintShit Thank you for your guidance; I attempted to follow your advice. 
Unfortunately, Streisand doesn't have an app version in its user-agent.

M03ED:
> @SaintShit Thank you for your guidance; I attempted to follow your advice. 
> Unfortunately, Streisand doesn't have an app version in its user-agent.

You can add it like v2rayng

PR #899, opened 2024-03-28
Reduce subscription link length

fodhelper:
This PR reduces subscription link
from :
`https://www.example.com/sub/eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ0ZXN0LTAxIiwiYWNjZXNzIjoic3Vic2NyaXB0aW9uIiwiaWF0IjoxNzA5MjE1MTUzfQ.omAlMh79ff1f5Btz9Ygt4JR9Gv4ZFfxtrm0-hoUkuew`
to
`https://www.example.com/sub/eyJzdWIiOiJ0ZXN0LTAxIiwiYWNjZXNzIjoic3ViIiwiaWF0IjoxNzA5MjE1MTUzfQ.omAlMh79ff1f5Btz9Ygt4JR9Gv4ZFfxtrm0-hoUkuew`

difference is 51 characters

old links will still work with no problem

### DIFF
--- a/app/utils/jwt.py
+++ b/app/utils/jwt.py
@@ -41,17 +41,22 @@ def get_admin_payload(token: str) -> Union[dict, None]:
 
 
 def create_subscription_token(username: str) -> str:
-    data = {"sub": username, "access": "subscription", "iat": datetime.utcnow()+timedelta(seconds=1)}
-    encoded_jwt = jwt.encode(data, get_secret_key(), algorithm="HS256")
+    data = {"sub": username, "access": "sub", "iat": datetime.utcnow()+timedelta(seconds=1)}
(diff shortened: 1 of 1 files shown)


M03ED:
Still to long , not much different

fodhelper:
Current difference is 51 characters!
Do you have any idea with JWT to reduce it even more?

M03ED:
> Current difference is 51 characters! Do you have any idea with JWT to reduce it even more?

in mu opinion there is no different between long and short sub link because you never gonna type it , you just copy it or use qr code

fodhelper:
yes, it's not that much important, but it's not that much useless too
As you said this PR's result is not short enough, i made another one without JWT, i will share it soon

PR #914, opened 2024-04-03
fix subscription link timezone

fodhelper:
patch for https://github.com/Gozargah/Marzban/issues/912 & change timezone of sub link's timstamp to utc

For future updates i guess must
change datetime.utcnow() and datetime.utcfromtimestamp() to other functions, as this ones are deprecated
And many functions has no timezone(will use server's timezone), add UTC to all

### DIFF
--- a/app/utils/jwt.py
+++ b/app/utils/jwt.py
@@ -43,7 +43,7 @@ def get_admin_payload(token: str) -> Union[dict, None]:
 
 
 def create_subscription_token(username: str) -> str:
-    data = username + ',' + str(int(datetime.now().timestamp()))
+    data = username + ',' + str(int(datetime.utcnow().timestamp()))
     data_b64_str = b64encode(data.encode('utf-8'), altchars=b'-_').decode('utf-8').rstrip('=')
     data_b64_sign = b64encode(sha256((data_b64_str+get_secret_key()).encode('utf-8')).digest(), altchars=b'-_').decode('utf-8')[:10]
     data_final = data_b64_str + data_b64_sign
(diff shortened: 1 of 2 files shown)


fodhelper:
This will fix the subscription problem
but timezone problem is more deep in Marzban
for future, must change `datetime.utcnow()` and `datetime.now()` to `datetime.now(timezone.utc)`
and must change `datetime.utcfromtimestamp(timezone)` and `datetime.fromtimestamp(timestamp)` to `datetime.fromtimestamp(timestamp, timezone.utc)`

app/utils/jwt.py
```diff
@@ -58,7 +58,7 @@ def get_subscription_payload(token: str) -> Union[dict, None]:
         if "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9." in token:
```
M03ED, line 58: i think its better to use `startswith` instead of `in` for finding `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.`

app/utils/jwt.py
```diff
@@ -43,7 +43,7 @@ def get_admin_payload(token: str) -> Union[dict, None]:
 
 
 def create_subscription_token(username: str) -> str:
-    data = username + ',' + str(int(datetime.now().timestamp()))
+    data = username + ',' + str(round(datetime.utcnow().timestamp()))
```
M03ED, line 46: using datetime ( instead of timestamp ) its not better ?

fodhelper: Timestamp is short and JWT used the same thing, so i just did the same

mikeesierrah (approved):

mikeesierrah:
I have tested this pr and found no problem - the problem seems to be fix now

mikeesierrah:
also this PR fixed the problem with subscription link not showing date and usage in Hiddify next client - thank you and goodjob

realartin:
we need this please merge it

PR #947, opened 2024-04-16
UserResponse: admin field added

Alirezaja1384:
Details:
 - User responses ship with admin info now.
 - CRUD's user queries: Admin left join.


Alirezaja1384:
Fixes #910

M03ED:
nice pr but you returning all admin data 
in this case this can cause data leak , users can see there admin data with `"/sub/{token}/info" ` request
we just need username or id

Alirezaja1384:
And since the non-sudo admins have no access to other admins' users, there would not be any risk of data leak cause it's the data that they should/and already have access to.

xmohammad1:
> i think its better to have separated schema for /info request with less data some data like `note` `sub_updated_at` `sub_last_user_agent` `online_at` `excluded_inbounds` `inbounds` are unnecessary

can you pls check your telegram i massaged you a group link most of develpors are there and we can make Decision

Alirezaja1384:
Since I don't like the pieces of code I wrote in the last commit, I chose to explain the changes. I know there are much better ways to handle this, but as far as I know, all of them require changes in `User` and `UserResponse`. Since UserResponse is highly integrated in every single piece of code (Which I don't like, but it is), any change that affects it can lead to unexpected bugs. So I chose to handle it in a way that has minimal side effects. Any suggestion is welcome as usual.

SaintShit:
Hello, thanks for your efforts. just tested it, and it returns users list empty.
Screenshot_2024-07-03_00-25-03

Alirezaja1384:
> Hello, thanks for your efforts. just tested it, and it returns users list empty. Screenshot_2024-07-03_00-25-03

Sounds to be fixed now. The join led to the exclusion of users who didn't have a db admin.

SaintShit:
Thank you @Alirezaja1384, changes are great. I might change the way "admin" field is returning in future.

Issue #931, opened 2024-04-07
Not supported Unix domain socket listener


Willh92:
inbound config not supported Unix domain socket listener when not set port

M03ED:
This is what xray should do not marzban , marzban it's just a program to manage your users.
You can report this to xtls organization

Willh92:
> This is what xray should do not marzban , marzban it's just a program to manage your users. You can report this to xtls organization

Xray is supported, but marzban checking inbound config not ignored port
# port settings
            try:
                settings['port'] = inbound['port']               
            except KeyError:
                if not self._fallbacks_inbound:
                    raise ValueError(
                        f"port missing on {inbound['tag']}"
                        "\nset XRAY_FALLBACKS_INBOUND_TAG if you're using an inbound containing fallbacks"
                    )
                try:
                    settings['port'] = self._fallbacks_inbound['port']
                    settings['is_fallback'] = True
                except KeyError:
                    raise ValueError("fallbacks inbound doesn't have port")

M03ED:
@SaintShit

evkravchenko:
It's a strange have to specify the port for inbound if it listening the unix socket. In my opinion Marzban should allow xray inbound configuration without port but with socket.

M03ED:
> It's a strange have to specify the port for inbound if it listening the unix socket. In my opinion Marzban should allow xray inbound configuration without port but with socket.

can you send a inbound with socket ?

evkravchenko:
Of course
for example, my real inbound of reality, for which I was forced to specify an extra port number:
```
{
    "tag": "reality",
    "listen": "/run/xray/nginx.socket,0666",
    "port": 443,
    "protocol": "vless",
    "settings": {
        "clients": [],
        "decryption": "none"
    },
    "streamSettings": {
        "network": "tcp",
        "tcpSettings": {
            "acceptProxyProtocol": true
        },
        "security": "reality",
        "realitySettings": {
            "dest": "/run/nginx/xray.socket",
            "xver": 1,
            "serverNames": [
                "server.name.sni"
            ],
            "privateKey": "*******************************************",
            "shortIds": [
                ""
            ]
        }
    },
    "sniffing": {
        "enabled": true,
        "destOverride": [
            "http",
            "tls",
            "quic"
        ],
        "routeOnly": true
    }
}
```
Here reality works behind nginx

Willh92:
> Of course
> 
[...]

Issue #896, opened 2024-03-28
HTTPUpgrade and Early Data for Clash.Meta


fodhelper:
in Clash.Meta the are both supported

HTTPUpgrade :
WS + `v2ray-http-upgrade: true` = HTTPUpgrade
https://github.com/MetaCubeX/mihomo/blob/06b5121d9eaebad3a8f1ff52cb7fd9d58a321823/docs/config.yaml#L463

Early Data :
https://github.com/MetaCubeX/mihomo/blob/06b5121d9eaebad3a8f1ff52cb7fd9d58a321823/docs/config.yaml#L461

Issue #885, opened 2024-03-23
multi custom config for v2rayN & gRPC updates


fodhelper:
multi custom config added to v2rayN>6.39 https://github.com/2dust/v2rayN/commit/262466303bb147ec03cdb5c1f15ddb3592716b18

xray-core added support for gRPC Authority (Supported by v2rayNG>1.8.18 and v2rayN>6.39)

gRPC's serviceName must be url-encoded in share-links
https://github.com/XTLS/Xray-core/discussions/716#:~:text=an%20empty%20string.-,4.3.13,-(gRPC)%20authority

M03ED:
these commits are for 12 hour ago and latest version of v2rayn (6.39) its for 2 weeks ago , you need to wait for new version.
Please be patient and do not rush

realartin:
> these commits are for 12 hour ago and latest version of v2rayn (6.39) its for 2 weeks ago , you need to wait for new version. Please be patient and do not rush

we need this please :) http upgrade

Issue #845, opened 2024-03-04
Telegram bot doesn't work properly


mahdim10:
Hello,
First of all, I would like to thanks all Marzban developers. However, there seems to be an issue with the Telegram bot not sending notifications for users who are about to expire. This problem mostly occurs at 00:00 when multiple users are reaching their expiration time. While it successfully notifies about some users, as indicated in the screenshot, it fails to notify all of them.
Im using the 0.4.9 version.
![photo_2024-03-04_15-28-02](https://github.com/Gozargah/Marzban/assets/30717578/1cf379b2-2a35-4baf-840f-a3b0e9d2b115)
![photo_2024-03-04_15-28-06](https://github.com/Gozargah/Marzban/assets/30717578/204318c6-1827-4f96-bbce-755dad68577f)

M03ED:
i have the same bug and i couldn't find why this is happening , if you find why this is happening send it here

fodhelper:
seems like rate limit of telegram api

M03ED:
> seems like rate limit of telegram api

nope its not that because users in begining of list will be skiped not last ones

fodhelper:
maybe should change
https://github.com/Gozargah/Marzban/blob/b2ad111eaf7d802e9c747ff34710a407ba3d3319/app/telegram/handlers/report.py#L25-L26

to
```
        except Exception as e:
            logger.error(e)
```
to see what the error is?

fodhelper:
I guess the problem could be this
https://github.com/Gozargah/Marzban/blob/1949717c0a622c20705f998e8e609263d660b269/app/utils/system.py#L119
It's a global variable, it can cause problems for urllib3 ipv6 connections (used in pyTelegramBotAPI too)
I have changed it in https://github.com/Gozargah/Marzban/pull/915 - maybe helps

PR #961, opened 2024-04-28
some fix and small features 

M03ED:
fix: 
- report change status temporary #845
- accept unix domain #931
- days left for expired users #903 
- fragment field in streisand #924 #827
- early data for clash #896 
- kcp for v2ray json
- quic for v2ray json #886 

feat: 
- add spiderx #957 
- add multi mode for grpc #885 #641
- add json config for v2rayn #659

refactor: 
- V2rayShareLink class usage
- remove external if else in share.py


fodhelper:
gRPC is changes in latest versions of Xray-core, check this : https://github.com/XTLS/Xray-core/pull/1815

There is 3 different type of gRPC server configs :
gRPC Gun (Supported on all clients)
gRPC Multi (Supported only on Xray-core based clients)
gRPC 2 in 1 Inbound: Both types in 1 inbound with Custom Path, for this one Marzban should parse Inbound's serviceName and use Multi for Xray based clients, and Gun for other clients

M03ED:
> gRPC is changes in latest versions of Xray-core, check this : [XTLS/Xray-core#1815](https://github.com/XTLS/Xray-core/pull/1815)

ok so we can have this type of grpc with `"network": "grpc"` or have normal grpc with `"network": "gun"` right ?

fodhelper:
network 'gun' and network 'grpc' are same thing, one transport with multiple aliases

M03ED:
> gRPC is changes in latest versions of Xray-core, check this : [XTLS/Xray-core#1815](https://github.com/XTLS/Xray-core/pull/1815)
> 
> There is 3 different type of gRPC server configs : gRPC Gun (Supported on all clients) gRPC Multi (Supported only on Xray-core based clients) gRPC 2 in 1 Inbound: Both types in 1 inbound with Custom Path, for this one Marzban should parse Inbound's serviceName and use Multi for Xray based clients, and Gun for other clients
> 
> And network name for gRPC can be 'grpc' or 'gun' other network names have aliases too. check this : https://github.com/XTLS/Xray-core/blob/51504c624c4aa9c093cf607cfca16e9548743af0/infra/conf/transport_internet.go#L631

done.

fodhelper:
Please revert "feat: support gRPC Custom Path", it's wrong

We must check if path is started with "/" or not
If it's not started with "/" then we must send path like before to all clients

But if it's started with "/" we must split it in serviceName and streamName(Gun) and streamName(Multi)
Example path = "/custom/path/GunCustomPath|MultiCustomPath"
  serviceName = custom/path
  streamName(Gun) = GunCustomPath
  streamName(Multi) = MultiCustomPath

now for Xray based clients we must send the full path ("/custom/path/GunCustomPath|MultiCustomPath")
in sharelink, this path must get urlEncoded
https://github.com/XTLS/Xray-core/releases/tag/v1.8.9
But for v2ray json config i think there is no need to urlEncode the path (must test it)

And when user uses this type of Path for gRPC, it does not mean MultiMode is Enabled, it could be disabled
This type of path only means we use custom streamName

For Clash.Meta and Sing-Box Gun mode will only work if streamName(Gun) is euqal to "Tun"
And we must send path to them like old type of configs (without "/" and without /streamName)

fodhelper:
You can add randomUserAgent function to the new function file

```python
import random

# Source: https://cdn.jsdelivr.net/gh/microlinkhq/top-user-agents@master/src/index.json
UserAgentList = [
  "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36",
  "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36",
  "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36"
]

def genRandUserAgent():
  return random.choice(UserAgentList)

print(genRandUserAgent())
```
You can add all 100 user agents from the url and update user agent list later too, but making many requests to jsdelivr cdn is not good, they may have or add a rate limit or block our server's ip

gRPC supports custom user-agent too, with user agent and authority of grpc we can use it with no tls and remain undetected

M03ED:
> Please revert "feat: support gRPC Custom Path", it's wrong
> 
> We must check if path is started with "/" or not If it's not started with "/" then we must send path like before to all clients
> 
> But if it's started with "/" we must split it in serviceName and streamName(Gun) and streamName(Multi) Example path = "/custom/path/GunCustomPath|MultiCustomPath" serviceName = custom/path streamName(Gun) = GunCustomPath streamName(Multi) = MultiCustomPath
> 
> now for Xray based clients we must send the full path ("/custom/path/GunCustomPath|MultiCustomPath") in sharelink, this path must get urlEncoded https://github.com/XTLS/Xray-core/releases/tag/v1.8.9 But for v2ray json config i think there is no need to urlEncode the path (must test it)
> 
> And when user uses this type of Path for gRPC, it does not mean MultiMode is Enabled, it could be disabled This type of path only means we use custom streamName
> 
> For Clash.Meta and Sing-Box Gun mode will only work if streamName(Gun) is euqal to "Tun" And we must send path to them like old type of configs (without the first "/" and without /streamName)

this is what they provide as a sample 

server side :`"serviceName": "/my/sample/path/customTun|customTunMulti"`
client without multi:
```
"serviceName": "/my/sample/path/customTun",
"multiMode": false
```
client with multi tun:
```
"serviceName": "/my/sample/path/customTunMulti",
"multiMode": true
```
if you don't want to use multi mode without conflict with not xray based client you can just use normal grpc 
we already sending path like before , only thing we do its separating grpc path for xray based clients and other ones
but i cant change that `custom` to be really custom

i'm not sure what you what you mean by urlEncode , we have different type of that

fodhelper:
> but i cant change that `custom` to be really custom

check this :

```python3
def get_grpc_gun(input: str):
    if not input.startswith("/"):
        return input
    
    servicename = input.rsplit("/", 1)[0]
    streamname = input.rsplit("/", 1)[1].split("|")[0]
    
    if streamname == "Tun":
        return servicename[1:]
    
    return "%s%s%s" % (servicename, "/", streamname)

def get_grpc_multi(input: str):
    if not input.startswith("/"):
        return input
    
    servicename = input.rsplit("/", 1)[0]
    streamname = input.rsplit("/", 1)[1].split("|")[1]

    return "%s%s%s" % (servicename, "/", streamname)

print(get_grpc_gun("oldPath"))
print(get_grpc_gun("/newPath/Tun|MultiCustomPath"))
print(get_grpc_gun("/new/path/GunCustomPath|MultiCustomPath"))

print(get_grpc_multi("oldPath"))
print(get_grpc_multi("/new/path/GunCustomPath|MultiCustomPath"))
```

> i'm not sure what you what you mean by urlEncode , we have different type of that

In xray-core tag 1.8.9 written `encodeURIComponent`
https://github.com/XTLS/Xray-core/releases/tag/v1.8.9
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent
so it must be

```python3
from urllib.parse import quote

print(quote("/my/path", safe="-_.!~*'()"))
```

M03ED:
so for not xray based clients we use get_grpc_gun and for xray based if multimode was enable we use get_grpc_multi?

M03ED:
and i can't find any user agent option for grpc

fodhelper:
> so for not xray based clients we use get_grpc_gun and for xray based if multimode was enable we use get_grpc_multi?

exactly
If multiMode is disabled >> get_grpc_gun (for all clients)
if multiMode is enabled >> get_grpc_multi (for xray based clients) and get_grpc_gun (for other clients)

> and i can't find any user agent option for grpc

They added it in https://github.com/XTLS/Xray-core/pull/1790/files
But they didn't added it in docs because it's useless! as no browser uses gRPC so using a user-agent of browser is useless
no \*ray client supports it too, so we must leave it for now

M03ED:
i fix grpc and add encode for service name start with /

M03ED:
@SaintShit

SaintShit:
thank you all great

PR #964, opened 2024-04-29
Fix v2ray-custom userLevel

fodhelper:
For policy section of v2rayNG we need to use a userLevel
i just exported an config from v2rayNG, it's default userlevel was 8 so i used 8 for default userLevel
image

### DIFF
--- a/app/subscription/v2ray.py
+++ b/app/subscription/v2ray.py
@@ -423,6 +423,7 @@ class V2rayJsonConfig(str):
         users["alterId"] = 0
         users["email"] = "https://gozargah.github.io/marzban/"
         users["security"] = "auto"
+        users["level"] = 8
         vnext["users"] = [users]
 
         return [vnext]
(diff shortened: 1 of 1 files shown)


M03ED:
I tried v2rayn and for vless it didn't use level , and for trojan use level 1

fodhelper:
Yes, i have exported a client config from v2rayN and it does not have it (even when statistics is enabled)
But in v2rayNG it does exist and level is 8 for all configs, it is used for Policy section when speed display is enabled from global settings

I wish there was separated template for v2rayNG and v2rayN, so i can disable logging for v2rayNG but keep it for v2rayN, and maybe some other changes

M03ED:
I test speed display enabled with level 8 but download and upload speed is 0

fodhelper:
Yes, that's because API inbound and settings is missing from v2rayNG exported config
So problem was not userLevel, but still Marzban must set same userLevel for all outbounds, or leave it to default 0 for trojan too

PR #969, opened 2024-05-03
Users auto-delete implemented.

Alirezaja1384:
TODO:
 - Send notifications


xmohammad1:
thats really good

Alirezaja1384:
Easier than expected :)

P.S. I didn't have time to test the notifications, but I don't see any reason for them not working.

M03ED:
thanks for this pr
first of all can you explain we have 3 migration here ? all of these can be one migration 
and for the next you are making new logger for just single job, marzban already have a logger , you can just use that

Alirezaja1384:
> first of all can you explain we have 3 migration here ? all of these can be one migration

Ok, I'm convinced. 3 is too much. I'll merge them :)

> and for the next you are making new logger for just single job, marzban already have a logger , you can just use that

I was in a hurry and didn't see it. thanks for your suggestion.

Alirezaja1384:
Sounds like I'm late to the party. I just merged the migrations :/

PR #993, opened 2024-05-18
regex fix

fodhelper:
In the freedom outbound, for fragment, you can set a single digit or a range
currently Marzban forces user to use a range, but i have changed it to accept single digit too


SaintShit:
thanks

PR #1019, opened 2024-05-29
fix: :bug: Disabled Clash Meta Flow for non-TLS configs

mahanmi:
non-TLS configs doesn't support flow and the config won't work if flow has been set so , i disabled it for non-TLS configs in clash-meta file (it was disabled for other clients)

### DIFF
--- a/app/utils/share.py
+++ b/app/utils/share.py
@@ -440,7 +440,7 @@ class ClashMetaConfiguration(ClashConfiguration):
         if inbound['protocol'] == 'vless':
             node['uuid'] = settings['id']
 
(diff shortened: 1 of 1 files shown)


M03ED:
its should be `inbound['tls'] != 'none'`
also your branch is old and needs to be updated , share.py file has been separated to multiple file in dev branch

mahanmi:
I'm so sorry about that my fork was for a year ago I'll close this PR and make a New one

mahanmi:
Shouldn't it doesn't have flow because of this line?
https://github.com/Gozargah/Marzban/blob/16ed7ec1bcd73a08248781f30d733685b8f7970a/app/xray/config.py#L311

PR #1031, opened 2024-06-04
feat(subscription): notify user when account expired

ImMohammad20000:
Now  you can specify a inbound for `NOTICE_INBOUND` in `.env` file 

When a user's subscription expires, remove all protocols from the subscription and leave one “YOUR SUBSCRIPTION EXPIRED” you can set your own text in hosts settings.

[feature requested here](https://github.com/Gozargah/Marzban/discussions/985#discussioncomment-9500460)


M03ED:
this is a good option but the way of implementation is not optimized , imagine we need to do this check every time refresh main page and we need to do the check for 30 user

mikeesierrah:
Can’t we consider an alternative approach? For instance, instead of disabling an account’s inbound and adding an inbound as a notifier, we could implement a system that:

Disables subscription updates when a client’s account expires, is limited, or deactivated.
Sends a simple notification message when the client updates their subscription.
Here’s what I mean:

When the client’s subscription status changes to expired, limited, or deactivated, the panel disables all subscription updates for that client.
When the client updates their subscription, the panel responds using a different inbound template.
This method eliminates the need to disable each protocol individually and then enable a specific one. The logic would be:

```
if user_status in ['limited', 'expired']:
    respond_with('limited.json')
else:
    respond_with('default.json')
```
This approach avoids the hassle of disabling and enabling inbounds and, in my opinion, is much more efficient.

fodhelper:
Why there is no separated messages for Expired, Limited and Disabled accounts?

and creating too many Inbounds will decrease Xray-core performance and waste resources
i think must config Marzban to not include this Fake Inbounds to Xray server config

ImMohammad20000:
> Why there is no separated messages for Expired, Limited and Disabled accounts?

separating each one needed a separate inbound and i don't want to make to many inbound for this, it decrease Xray-core performance

ImMohammad20000:
> this is a good option but the way of implementation is not optimized , imagine we need to do this check every time refresh main page and we need to do the check for 30 user

any better idea tell me i'll make it better

fodhelper:
A more simple way is to have a Env Variable to remove all Hosts except the ones with '0.0.0.0' as Address (for non-active users)
a combination of this env variable and STATUS_TEXT (https://github.com/Gozargah/Marzban/pull/898) can done the job

erfjab:
We must find another way.

PR #1033, opened 2024-06-04
Fix Uvicorn settings when config.DEBUG is True

fodhelper:
Am i wrong?

### DIFF
--- a/main.py
+++ b/main.py
@@ -16,9 +16,9 @@ if __name__ == "__main__":
     try:
         uvicorn.run(
             "main:app",
-            host=('0.0.0.0' if DEBUG else UVICORN_HOST),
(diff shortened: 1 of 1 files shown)


M03ED:
yeap this is wrong 
these options are for when we need to use debug mode 
you just ignore it

fodhelper:
My `UVICORN_HOST` is `127.0.0.1` and i don't want to expose it to web directly
why should it change to `0.0.0.0` by enabling DEBUG? i don't understand it

SaintShit:
I do test the app over all networks of machine while I'm on DEBUG mode. And this is set just because I'm too lazy :)) this suppose to be just for development, I don't see any wrong with it.

PR #1038, opened 2024-06-06
Remove enabling 'multiplex' in sing-box config when mux is enabled

oXIIIo:
Previously, enabling mux for v2rayng would also enable 'multiplex' in the sing-box configuration, causing the config to not work as intended. This change ensures that 'multiplex' is not enabled in sing-box config even if mux is enabled for v2rayng. Enabling and disabling mux can now be managed directly within the sing-box mux template.

### DIFF
--- a/app/subscription/singbox.py
+++ b/app/subscription/singbox.py
@@ -187,7 +187,6 @@ class SingBoxConfiguration(str):
             mux_json = json.loads(self.mux_template)
             mux_config = mux_json["sing-box"]
             config['multiplex'] = mux_config
-            config['multiplex']["enabled"] = mux_enable
 
         return config
 


M03ED:
i don't know why this changes applied but in previous version we were able to disable sing-box mux for all configs in mux template
@SaintShit

mikeesierrah:
lets Just comment the code out without breaking anything
line 185 to 190 is about mux getting enabled we can simply comment it out 

I also don't understand it was uncommented

M03ED:
> lets Just comment the code out without breaking anything line 185 to 190 is about mux getting enabled we can simply comment it out
> 
> I also don't understand it was uncommented

that change was unnecessary , sanit change it after that and make it enable for sing-box  16ed7ec1bcd73a08248781f30d733685b8f7970a

plasticgholam:
منکه دستی ادیتش زدم تا سینت از سفر برگرده تاییدش کنه :)

fodhelper:
@mikeesierrah You made an unnecessary change (https://github.com/Gozargah/Marzban/pull/1006) and ruined everything!
Mux was already disabled for Sing-Box by default but you could enable it using Mux Template (https://github.com/Gozargah/Marzban/pull/882)
Thank you for creating problems

@SaintShit Please revert mux settings back to https://github.com/Gozargah/Marzban/pull/882/files

mikeesierrah:
> @mikeesierrah You made an unnecessary change (https://github.com/Gozargah/Marzban/pull/1006) and ruined everything!
> Mux was already disabled for Sing-Box by default but you could enable it using Mux Template (https://github.com/Gozargah/Marzban/pull/882)
> Thank you for creating problems
> 
> @SaintShit Please revert mux settings back to https://github.com/Gozargah/Marzban/pull/882/files

Ah, I see. We’ve been adding a disabled mux template to every configuration, even though it’s unnecessary. Our code is so robust that it requires confirmation from both the front end and the template to prevent accidental mux activation in the singbox. 
Because, of course, it’s entirely reasonable to have an elaborate process just to ensure we don’t enable mux on singbox. It’s not like we’re making things unnecessarily complicated for no good reason, right?

I’ve corrected it for you. Now, it doesn’t need double confirmation, nor does it add a redundant mux-disabled template to every singbox configuration, which was causing a mess with the Hiddify client. 

Also, if you check my PR, **you’ll notice I specifically commented out** the code and left a note stating that this code should not be uncommented unless the Marzban core supports singbox.




Please direct your inquiries to whoever uncommented that section, instead of attributing the error to me.

M03ED:
> > @mikeesierrah You made an unnecessary change (https://github.com/Gozargah/Marzban/pull/1006) and ruined everything!
> > Mux was already disabled for Sing-Box by default but you could enable it using Mux Template (https://github.com/Gozargah/Marzban/pull/882)
> > Thank you for creating problems
> > 
> > @SaintShit Please revert mux settings back to https://github.com/Gozargah/Marzban/pull/882/files
> 
> Ah, I see. We’ve been adding a disabled mux template to every configuration, even though it’s unnecessary. Our code is so robust that it requires confirmation from both the front end and the template to prevent accidental mux activation in the singbox. 
> Because, of course, it’s entirely reasonable to have an elaborate process just to ensure we don’t enable mux on singbox. It’s not like we’re making things unnecessarily complicated for no good reason, right?
> 
> I’ve corrected it for you. Now, it doesn’t need double confirmation, nor does it add a redundant mux-disabled template to every singbox configuration, which was causing a mess with the Hiddify client. 
> 
> Also, if you check my PR, **you’ll notice I specifically commented out** the code and left a note stating that this code should not be uncommented unless the Marzban core supports singbox.
> 
> 
> 
> 
> Please direct your inquiries to whoever uncommented that section, instead of attributing the error to me.

First rule of programing:
If it's work , don't change it

M03ED:
done in #1071

Issue #1052, opened 2024-06-18
Support FlClash client app


wikm360:
FlClash (uses clash meta core) send request with "clash.meta" user agent , but marzban in response send clash template ( without vless ) instead of ClashMeta template .

FlClash github : https://github.com/chen08209/FlClash

Thanks...

PR #1065, opened 2024-06-30
fix: add support for FlClash client app and resolve issue #1052

oXIIIo:
Updated the user agent regex to include support for the FlClash client. This resolves issue #1052.

### DIFF
--- a/app/views/subscription.py
+++ b/app/views/subscription.py
@@ -79,7 +79,7 @@ def user_subscription(token: str,
 
     crud.update_user_sub(db, dbuser, user_agent)
 
-    if re.match('^([Cc]lash-verge|[Cc]lash-?[Mm]eta)', user_agent):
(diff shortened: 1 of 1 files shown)


fodhelper:
I guess this PR will not solve the problem, because user-agent of FlClash is :
`FlClash/v0.8.33 clash-verge/v1.6.6 Platform/android`
I suggest this RegEx :
`([Cc]lash-[Vv]erge|[Cc]lash[-\.]?[Mm]eta|[Mm]ihomo)`
Mihomo is the new name of Clash.Meta, some projects uses it like https://github.com/YukiCoco/ToMoon
or maybe this
`re.match('(clash-verge|clash[-\.]?meta|mihomo)', user_agent, re.IGNORECASE)`
we can add FlClash itself too, maybe they remove clash-verge from their user-agent later?

PR #1066, opened 2024-07-01
v2ray-json conf only for 1.8.18+ versions of v2rayNG

fodhelper:
1.8.16 and 1.8.17 versions of v2rayNG is not able to parse json configs correctly, they can't parse early data parameter and with some json templates they can't even connect to configs
so it's better to just send them old type v2ray configs


m0x61h0x64i:
ایده من بود ولی دلیل کافی نداشتم :/

PR #1072, opened 2024-07-03
add randomization of sub configs

erfjab:
You can randomize the sub configs from the env. It mixes up the configs, so with each sub update, the configs get rearranged differently.


erfjab:
https://github.com/Gozargah/Marzban/assets/143827987/a0f2d57b-ce85-45d0-8ccf-34f3ad20b5d0

M03ED:
can i ask what is use case of this ?

erfjab:
This feature is really useful when you have a lot of configs and your users only connect to the first one. If you're unhappy because your Shadowsocks and Trojan configs are at the bottom of the list, this will help. It's something none of the other panels have.

erfjab:
A lot of friends in the Marzban group were requesting this feature. It's optional and the code is really lightweight.

PR #1076, opened 2024-07-03
some new features

ImMohammad20000:
- Add support for splithttp
- Add authority support for gRPC
- Add random user-agent for gRPC


iambabyninja:
Big thanx 4 your work ❤️🤝♾️

M03ED:
grpc is not supported by any browser and using user-agent can make it so easy to detect
splithttp implantation it wrong and there is some missing fields ( i'm gonna do it in next few days )

SaintShit:
thanks that's great

Issue #1034, opened 2024-06-05
ALPN bug in dev branch


develfishere:
Hello dear Gozargah team and Thank you very much for you help to bypass internet restrictions.

In latest version of dev branch, when we set ALPN to the combination of http/1.1 and h2, the panel return it as a string (i.e, "h2, http/1.1" ) into json config and does not split it as separate values for the ALPN array of json config. 
The ALPN key in the json config would be as:
"alpn": [
    "h2,http/1.1"
],
instead of:
"alpn": [
    "h2" ,
    "http/1.1"
],
Also, setting h3 (or any combination with h3) in ALPN section, makes the host settings not being saved.

fodhelper:
I have tried to use h3 as ALPN and failed
Here is logs :
```
INFO:     127.0.0.1:0 - "PUT /api/hosts HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1900, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
  File "/usr/local/lib/python3.10/site-packages/pymysql/cursors.py", line 158, in execute
    result = self._query(query)
  File "/usr/local/lib/python3.10/site-packages/pymysql/cursors.py", line 325, in _query
    conn.query(q)
  File "/usr/local/lib/python3.10/site-packages/pymysql/connections.py", line 549, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
  File "/usr/local/lib/python3.10/site-packages/pymysql/connections.py", line 779, in _read_query_result
    result.read()
  File "/usr/local/lib/python3.10/site-packages/pymysql/connections.py", line 1157, in read
    first_packet = self.connection._read_packet()
  File "/usr/local/lib/python3.10/site-packages/pymysql/connections.py", line 729, in _read_packet
    packet.raise_for_error()
  File "/usr/local/lib/python3.10/site-packages/pymysql/protocol.py", line 221, in raise_for_error
    err.raise_mysql_exception(self._data)
  File "/usr/local/lib/python3.10/site-packages/pymysql/err.py", line 143, in raise_mysql_exception
    raise errorclass(errno, errval)
[...]

PR #1077, opened 2024-07-03
fix: splithttp and alpn

M03ED:
- add missing parameters for splithttp #1076
- fix alpn bug #1034

### DIFF
--- a/app/db/migrations/versions/305943d779c4_add_h3_to_alpn_enum.py
+++ b/app/db/migrations/versions/305943d779c4_add_h3_to_alpn_enum.py
@@ -1,0 +1,118 @@
+"""add h3 to alpn enum
+
+Revision ID: 305943d779c4
+Revises: 31f92220c0d0
+Create Date: 2024-07-03 19:27:15.282711
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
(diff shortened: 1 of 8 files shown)


M03ED:
- fix for #798 in commit [e22bf9b](https://github.com/Gozargah/Marzban/pull/1077/commits/e22bf9b0284f0a519c899bd08c010e62de50f07b)

iambabyninja:
My dear friend, big thanx 4 your work.

Default values for this parameters:

func (c *Config) GetNormalizedMaxConcurrentUploads() int32 {
	if c.MaxConcurrentUploads == 0 {
		return 10
	}

	return c.MaxConcurrentUploads
}

func (c *Config) GetNormalizedMaxUploadSize() int32 {
	if c.MaxUploadSize == 0 {
		return 1000000
	}

	return c.MaxUploadSize
}

M03ED:
> The value should be set in bytes, accordingly, the standard recommended chunk size of 1MB should be set to approximately 1,000,000 bytes.

done.

fodhelper:
Please change initial_windows_size to 35536 to make cloudflare grpc configs work
image
https://xtls.github.io/en/config/transports/grpc.html#grpcobject

PR #1090, opened 2024-07-05
Fix grpc v2ray json

fodhelper:


fodhelper:
not tested yet

There is a bug for QUIC, maybe exist for KCP too
`quicSettings["header"]["type"] = header`

we should parse v2rayNG version and check if users' app does not support a specific transport, we should not send that config in json, because if even 1 config be unknown to the client, it will fail the subscription update process

we must also should not send unsupported sing-box configs to it, because sing-box update will fail (something like QUIC without TLS or or with quic with obfuscation)

ImMohammad20000:
can you send your document about this change

fodhelper:
new versions of Xray-core need full path for gRPC Path or Custom Path

Old clients will not url decode gRPC path, encoding them will just make problems if `/` exist in the path, but if new clients try to url decode the path, `/` will be still `/`, and i guess existing `/` in the path don't make any problem in parsing urls for any client app

for v2rayN / v2rayNG outbound tag is "proxy" exactly like sing-box and clash.meta (they all need unique tag for selected proxy, to use them in dns and route rules)
it should not be the config name, config name is `remarks`

Issue #1108, opened 2024-07-15
Add node error monitoring


erfjab:
Whenever "nodes" get an error and exit the Connect mode, it will be notified.

PR #1117, opened 2024-07-16
add node error monitoring

erfjab:
With this feature, you will be notified whenever there is a problem connecting the nodes.
#1108

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -64,6 +64,10 @@ UVICORN_PORT = 8000
 # USERS_AUTODELETE_DAYS = -1
 # USER_AUTODELETE_INCLUDE_LIMITED_ACCOUNTS = false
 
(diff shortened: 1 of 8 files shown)


erfjab:
image
For example

M03ED:
in current implementation when node get error and stuck their we gonna get a spam message from marzban , every 10 second we receive new message 
i tried to write this before and i didn't find good way to avoid spam

erfjab:
> in current implementation when node get error and stuck their we gonna get a spam message from marzban , every 10 second we receive new message i tried to write this before and i didn't find good way to avoid spam

Of course, there is a way, but it is not very professional, here we can add that it will wait 60 seconds after every error, but we will be forced to ignore the scheduler error. 😅


image

erfjab:
I should write better

PR #1118, opened 2024-07-16
add External config to import into v2ray format sub

erfjab:
With this feature, you can automatically apply temporary V2Ray configurations to your `active users' subscriptions` (in `V2Ray format`) during connection outages, allowing them to stay connected until maintenance is completed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,6 +51,9 @@ UVICORN_PORT = 8000
 # SUB_UPDATE_INTERVAL = "12"
 # RANDOMIZE_SUBSCRIPTION_CONFIGS = True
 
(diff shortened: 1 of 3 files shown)


erfjab:
#1118

M03ED:
what about sing-box and clash users ?

erfjab:
> what about sing-box and clash users ?

For now, I tried to make it as simple as possible. For users other than v2ray, we need to parse the input configurations, which makes the code complex and heavy. If you want us to do that, I will do it.

SaintShit:
this is just a temporary solution. I'll merge it for now, and not include to readme. once we implement a better solution, we remove it.

Issue #1121, opened 2024-07-16
Generate Subscription QR Separately in Telegram Bot 


iamtheted:
Telegram has got limits for bots to send multiple files at a time so when we tap on "Links" button in Marzban's Telegram bot it doesn't send the QRs if they're too many.

So I suggest we allow the subscription link QR to be generated separately. 

Right now, I need to take a screenshot from subscription link QR in the panel which is a little frustrating. 

Please take this into consideration if possible. 💙

iamtheted:
Thank you for the PR mate @erfjab 
I'll close this issue then.

PR #1123, opened 2024-07-17
fix Generate Subscription QR Separately in Telegram Bot #1121

erfjab:
image

i fix this #1121


erfjab:
The problem that the links section had, if the length of the text of the links was too long, it could not be sent, which was temporarily solved by limiting the length of the text of the links.

iamtheted:
Perfect! Thank you so much bro! 💙 
I already tested it, and it works! 
No matter how many configs you have, you can get the subscription link QR separately.

SaintShit:
Thank you for helping us @erfjab @iamtheted

PR #1124, opened 2024-07-17
fix: update user_expire_step to accept days instead of date

erfjab:
With this feature, you don't need to enter the exact date anymore, just by entering the desired day, it will record the date itself.

### DIFF
--- a/app/telegram/handlers/admin.py
+++ b/app/telegram/handlers/admin.py
@@ -4,7 +4,7 @@ import os
 import random
 import re
 import string
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import qrcode
 import sqlalchemy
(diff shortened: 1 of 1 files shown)


ImMohammad20000:
You can use d|m|y to do this now
5d => 5 days
2m => 2 month
1y => 1 year

Its better than only use days for expire date

app/telegram/handlers/admin.py
```diff
@@ -382,7 +382,7 @@ def edit_user_command(call: types.CallbackQuery):
     elif action == "expire":
         msg = bot.send_message(
             call.message.chat.id,
-            '⬆️ Enter Expire Date (YYYY-MM-DD)\nOr You Can Use Regex Symbol: ^[0-9]{1,3}(M|D) :\n⚠️ Send 0 for never expire.',
```
M03ED, line 385: you could just use regex , no need for this change

M03ED:
no need for this pr , this is already available

erfjab:
Yes, that's right. I hadn't paid attention to him.

PR #1130, opened 2024-07-17
Update: Change Custom config email

Kiya6955:
- Custom config email changed from `https://gozargah.github.io/marzban` to `t@t.com`.

### DIFF
--- a/app/subscription/v2ray.py
+++ b/app/subscription/v2ray.py
@@ -650,7 +650,7 @@ class V2rayJsonConfig(str):
         vnext["port"] = port
         users["id"] = id
         users["alterId"] = 0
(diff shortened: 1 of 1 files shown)


M03ED:
this change have no affect and this is like a signature for marzban

M03ED:
> but it's better if we could customize it, maybe we don't want others to know what panel we're using.

there is much simpler way to find out what panel you use
for a user who wants to find it it's so simple and this is not gonna stop him
and for those users don't have knowledge about xray and anti censorship tools it doesn't matter what panel you use

M03ED:
I don't see any reason this pr remain open

PR #1133, opened 2024-07-18
refactor(subscription)

M03ED:
- simpler code for v2ray-json
- fix for #1127

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -276,7 +276,7 @@ class UserResponse(User):
     def validate_links(cls, v, values, **kwargs):
         if not v:
             return generate_v2ray_links(
-                values.get("proxies", {}), values.get("inbounds", {}), extra_data=values
+                values.get("proxies", {}), values.get("inbounds", {}), extra_data=values, reverse=False,
             )
         return v
 
--- a/app/subscription/clash.py
+++ b/app/subscription/clash.py
@@ -39,7 +39,9 @@ class ClashConfiguration(object):
         else:
             self.grpc_user_agent_data = []
 
-    def render(self):
+    def render(self, reverse=False):
+        if reverse:
+            self.data['proxies'].reverse()
         return yaml.dump(
             yaml.load(
                 render_template(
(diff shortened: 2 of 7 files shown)


ImMohammad20000:
I think it is better to have reverse as environment variable because maybe they change the way apps loads configs in future

M03ED:
> I think it is better to have reverse as environment variable because maybe they change the way apps loads configs in future

Should we add 11 different variable to env just for this ?

ImMohammad20000:
> > I think it is better to have reverse as environment variable because maybe they change the way apps loads configs in future
> 
> Should we add 11 different variable to env just for this ?

we have 6 config format

- v2ray
- v2ray-json
- clash
- clash-meta
- sing-box
- outline

and as far as i know currently all xray based apps do the reversing for json format

we need one variable per format for this

M03ED:
> we have 6 config format
> 
> - v2ray
> - v2ray-json
> - clash
> - clash-meta
> - sing-box
> - outline
> 
> and as far as i know currently all xray based apps do the reversing for json format
> 
> we need one variable per format for this

All appliances are ok 
Only problem is v2rayng , it will insert config in reverse and it will start from end of array
Because of this we decide to insert them in reverse order to be ok in v2rayng 

Also this is not about config type , this is about user agent

ImMohammad20000:
> > we have 6 config format
> > 
> > * v2ray
> > * v2ray-json
> > * clash
> > * clash-meta
> > * sing-box
> > * outline
> > 
> > and as far as i know currently all xray based apps do the reversing for json format
> > we need one variable per format for this
> 
> All appliances are ok Only problem is v2rayng , it will insert config in reverse and it will start from end of array Because of this we decide to insert them in reverse order to be ok in v2rayng
> 
> Also this is not about config type , this is about user agent

v2rayn and Streisand do this too

M03ED:
> v2rayn and Streisand do this too

No they don't, they show it in reverse because marzban send it in reverse

PR #1136, opened 2024-07-18
feat: add on_hold status in tgbot and refactore random username #1125

erfjab:
I have added the `on_hold` status to the Marzban Telegram bot so that users do not have to rely on secondary solutions like Holderbot or ... (unless they prefer to). Here are the tasks I completed:

- [x] **Added:** Ability to select on_hold status when creating a user
- [x] **Added:** Count of on_hold users in system info, subtracting them from inactive users
- [x] **Refactored:** User statistics text
- [x] **Refactored:** Random username generate
- [x] **Fixed:** User count in the user list
- [x] **Fixed:** Icon for on_hold users

### DIFF
--- a/app/telegram/handlers/admin.py
+++ b/app/telegram/handlers/admin.py
@@ -4,7 +4,7 @@ import os
 import random
 import re
 import string
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import qrcode
 import sqlalchemy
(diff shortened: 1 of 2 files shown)


SaintShit:
Sadly, I wasn't aware that on_hold is not included to bot. big thanks.

PR #1137, opened 2024-07-19
On Hold Property Added to UserDialog | Usage Coefficient Added to NodesModal

x0sina:
Coefficient
Screenshot 2024-07-19 224321


SaintShit:
Nice! we've needed this for a long time.
Can you please make the "Expire duration" field required when the On_hold is checked? otherwise, this error will throw

image

and it's not catched.

M03ED:
also this change is on master branch not dev branch

x0sina:
Thanks for reviewing my pull request
Expire duration is now required in UserDialog when trying to create or edit a user on on_hold' status.

M03ED:
Thanks for this, but we should not be able to change user status to on_hold when we edit the user ?

PR #1139, opened 2024-07-19
Fix grpc quic

fodhelper:


app/subscription/singbox.py
```diff
@@ -256,6 +256,11 @@ def add(self, remark: str, address: str, inbound: dict, settings: dict):
         if net in ("kcp", "splithttp"):
             return
 
+        if net == "quic" and inbound["header_type"] != 'none':
+            return
+
+        if
```
SaintShit, line 262: what's this if? it's always better to have project in your code space and run a test before making PR. (as we don't have test actions sadly)

fodhelper: Sorry, I have actually tested it, idk where this 'if' came from

fodhelper: As i don't use Git, this problems occur

PR #1146, opened 2024-07-19
Refactore: randomize sub configs

M03ED:


erfjab (changes requested):
You have used the randomization of inbounds , you should use the randomized hosts. Of course, if I am wrong, I would appreciate your guidance.

PR #1151, opened 2024-07-20
feat(telegramBot): Add onhold timeout

ImMohammad20000:


app/telegram/handlers/admin.py
```diff
@@ -1252,14 +1257,70 @@ def add_user_expire_step(message: types.Message, username: str, data_limit: int,
 
     schedule_delete_message(message.chat.id, message.id)
     cleanup_messages(message.chat.id)
+    if user_status == "onhold":
+        timeout_message = '⬆️ Enter timeout (YYYY-MM-DD)\nOr You Can Use Regex Symbol: ^[0-9]{1,3}(M|D) :\n⚠️ Send 0 for never expire.'
```
SaintShit, line 1261: 0 for never timeout, not expire

PR #1154, opened 2024-07-20
Enhance SNI and Host Selection Logic Based on List Lengths

amotlagh:
If the lengths of sni_list and req_host_list are the same, it selects a random index and uses the corresponding elements from both lists. If the lengths are different, it falls back to the original behavior of selecting random elements independently.

### DIFF
--- a/app/subscription/share.py
+++ b/app/subscription/share.py
@@ -303,16 +303,20 @@ def process_inbounds_and_tags(
             host_inbound = inbound.copy()
             for host in xray.hosts.get(tag, []):
                 sni = ""
-                sni_list = host["sni"] or inbound["sni"]
-                if sni_list:
-                    salt = secrets.token_hex(8)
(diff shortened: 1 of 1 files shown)


M03ED:
I submit this change to make pattern more random 
May I know the reason for this change?

erfjab:
> I submit this change to make pattern more random 
> May I know the reason for this change?


Some SNI require some host. When you want to use the Random Host and SNI feature, this feature is very useful.

erfjab (approved):
Nice 🤌

Mmdd93:
We use cdn config so host and sni in 90% of situations must be the same. when we try to use multiple subdomains in host and sni, the output is dissimilar ( unequal sni and host) so our cdn config does not work. first logic: We can only use multiple subdomains in host or sni, and the empty field (sni or host) should be filled as the same.
sec logic: if we have equal multiple subdomains in host and sni, they fill as equal subdomains in the output config

PR #1162, opened 2024-07-21
feat: add address list support in hosts

erfjab:
For traffic load balance and diversity in IP or address ranges of the configurations, it was necessary to list them.

**My Tasks:**

- [x] add address list support in hosts
- [x] support domain and ip


erfjab:
https://github.com/user-attachments/assets/e59ff3cd-e6aa-43b0-8f85-971db058f210

fodhelper:
Great, It's also good to support ports like "1005-1010,5000"
But needs too many changes, Front, Back and Database

erfjab:
> we already discuss about this , easier and more efficient way is using a domain that point to multiple ip or domain

We provide the option, now you can use it, I can use this. We can't force our preferences on anyone.

PR #1175, opened 2024-07-23
Prevent zooming on input focus in iOS

x0sina:
### Overview
This pull request adds a meta tag to the project to prevent zooming when focusing on an input field in iOS devices. This change addresses a common usability issue where the viewport zooms in when a user interacts with form fields, which can disrupt the user experience.

### Changes
- Added `<meta name="viewport" content="width=device-width, user-scalable=no" />` to the HTML `<head>` section.

### Reasoning
When users focus on an input field in iOS, the default behavior is to zoom in to provide a closer view of the text. While this can be helpful in some contexts, it often results in an awkward and jarring experience, especially in responsive web applications. By setting `user-scalable` to `no`, we prevent this automatic zooming behavior, ensuring a smoother and more consistent user experience across devices.

### Testing
- Verified that the meta tag is included in the HTML output.
- Tested on iOS devices to ensure the input fields no longer trigger zooming.
- Checked across various browsers to confirm no adverse effects on non-iOS devices.

### Notes
This change targets iOS-specific behavior and does not affect other aspects of the viewport settings.


M03ED:
i suggest be more careful in your pr's
marzban merge your pr 2 times and both of them cause a massive issue in frontend, i suggest you to fix last one then apply new changes

x0sina:
You can see the commits, Changes that applied after my PR made that issues

PR #1187, opened 2024-07-25
feat: Add migration to correct last_status_change for expired users

Alirezaja1384:
P.S. I don't know if handling only MySQL and SQLite is enough or not.


erfjab (commented):
Friends also use Mariadb, but the Mariadb and MySQL commands are similar.

SaintShit:
Hi, would you please explain what problem is solves?

Alirezaja1384:
> Hi, would you please explain what problem it solves?

Sure, the migration I wrote before sets that `last_status_change` to migration time, so if we suppose the expired should be deleted in 30 days, they will be deleted a month after the migration, not their expiration date. This migration sets the `last_status_change` to the user's expiration time.

ImMohammad20000:
what if user expire is null??

ImMohammad20000:
> what if user expire is null??

@Alirezaja1384

PR #1191, opened 2024-07-25
feat(subtitle): add ability to change subtitle to user's username

erfjab:
Some users have requested that in software like `Hiddify Next`, instead of the `Subscription title`, the username of the users should be displayed. I have added this feature as an `option`. If you write `username for the Subscription title`, each user's title will become their username; otherwise, whatever else you write will be displayed as the title.

- [x] feat: add ability to change subtitle to user's username

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -285,8 +285,9 @@ class UserResponse(User):
         if not v:
             salt = secrets.token_hex(8)
             url_prefix = (XRAY_SUBSCRIPTION_URL_PREFIX).replace('*', salt)
-            token = create_subscription_token(values["username"])
-            return f"{url_prefix}/{XRAY_SUBSCRIPTION_PATH}/{token}"
+            username = values["username"]
+            token = create_subscription_token(username)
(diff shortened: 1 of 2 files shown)


erfjab:
I made another commit to append the user's username with a hashtag at the end of the subscription link. What is the purpose of this?

1. Firstly and most importantly, the user's username becomes identifiable from the subscription link.
2. Secondly, some software (most notably v2rayng) can only recognize the subscription name in this manner.

erfjab:
image
image

ImMohammad20000:
This is not useful you can send username with host variables for user

erfjab:
> This is not useful you can send username with host variables for user

What if I want to send it in the header?

ImMohammad20000:
Why someone would send it in the header??

erfjab:
> Why someone would send it in the header??

Because it wants the user's username to be included in the title when importing subs.

xmohammad1:
> This is not useful you can send username with host variables for user

is not same as when you put it in host

PR #1237, opened 2024-08-10
increased fragment settings to 4 digits

ImMohammad20000:
for some reason 4 digit fragment settings can help to bypass tlshello fragment blocking


erfjab (approved):
Thank you. And Thank you for your help in updating the X-ray core fragment.

PR #1248, opened 2024-08-16
Fix(users): add ilike to get user

erfjab:
A small change for me, but a big change for tgbots. :)

### DIFF
--- a/app/db/crud.py
+++ b/app/db/crud.py
@@ -100,7 +100,7 @@ def get_user_queryset(db: Session) -> Query:
 
 
 def get_user(db: Session, username: str):
-    return get_user_queryset(db).filter(User.username == username).first()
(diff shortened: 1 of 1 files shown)


M03ED:
this is for finding exact username, if you want to find similar username use `get_users`

erfjab:
> this is for finding exact username, if you want to find similar username use `get_users`
> 

My goal was to be able to get a user's information regardless of whether it is upper or lower.

M03ED:
marzban database is sensitive to upper case and lower case 
you can have `user` and `User` at same time so this can cause lots of bugs
so i close this pr

erfjab:
> marzban database is sensitive to upper case and lower case 
> you can have `user` and `User` at same time so this can cause lots of bugs
> so i close this pr

Ok.

Issue #71, opened 2023-03-05
Change dashboard path


A3is:
hi
Please add this feature to change the address of the panel.

Current address :
Domain/dashboard

For example :
Domain/MyAddress

thank you ♥

alexeyp0708:
OK. What about the paths /api/, /statics/, /sub/?
They return the result {"detail":"Not Found"}.
It is enough to check these paths to increase suspicions about the Marzban service.

PR #1278, opened 2024-08-31
Feature -> configurable dashboard path

MR-MKZ:
> [!Note] 
> currently, It's just available for dev mode :/

- [x] #71 

image


erfjab:
Hi, good pr my friend. please fix your git.

erfjab:
> currently, It's just available for dev mode :/


What do you mean?

MR-MKZ:
> What do you mean?

I mean when user use docker for panel it doesn't work because it need to build panel frontend with new path and docker container doesn't have npm to build.
I tried to install node js on container but I couldn't even build the old image and container package manager returned ign when tried to install packages :/

.env.example
```diff
@@ -1,33 +1,36 @@
...
+# SUDO_PASSWORD="admin"
 
 # UVICORN_UDS: "/run/marzban.socket"
-# UVICORN_SSL_CERTFILE = "/var/lib/marzban/certs/example.com/fullchain.pem"
-# UVICORN_SSL_KEYFILE = "/var/lib/marzban/certs/example.com/key.pem"
+# UVICORN_SSL_CERTFILE="/var/lib/marzban/certs/example.com/fullchain.pem"
+# UVICORN_SSL_KEYFILE="/var/lib/marzban/certs/example.com/key.pem"
 
+# To change default dashboard path. Don't forget / at the start and the end of path.
+DASHBOARD_PATH=dashboard
```
M03ED, line 15: This should be a comment in example file, also you forgot to use / in start and end and " for string

build_dashboard.sh
```diff
@@ -1,3 +1,14 @@
-cd `dirname $0`/app/dashboard
-VITE_BASE_API=/api/ npm run build --if-present -- --outDir build --base '/dashboard/'
-cp ./build/index.html ./build/404.html
\ No newline at end of file
+#!/usr/bin/env bash
```
M03ED, line 1: Is it better to get path as input instead of reading env file again.

M03ED (changes requested):

.env.example
MR-MKZ: / at start and end will set automatically and ok I add " for string and make this env as a comment

build_dashboard.sh
MR-MKZ: Hmm, that's a good idea💡

erfjab:
You forget / 😅

MR-MKZ:
> You forget / 😅
Where?

erfjab:
> Where?

No, I read it again, it's OK.

MR-MKZ:
I found some bugs and I need to fix them before merge so I make this pr draft for a short time to fix bugs.

MR-MKZ:
Structure of this feature has some problems and I have no idea to how fix it

Issue #1279, opened 2024-09-01
Number Range for SplitHTTP Settings


fodhelper:
Marzban handles SplitHTTP settings very bad
for example when you set String(Range Number) for scMaxEachPostBytes and scMaxConcurrentPosts, Marzban set the same for maxUploadSize and maxConcurrentUploads
but v2rayNG expect a Int for maxUploadSize and maxConcurrentUploads, not a String, so subscription update will fail
https://github.com/2dust/v2rayNG/blob/9a9d315e627271e8fc21f5aa51a6c90262cbf73c/V2rayNG/app/src/main/kotlin/com/v2ray/ang/dto/V2rayConfig.kt#L243-L244

Issue #1277, opened 2024-08-31
Marzban overrides SplitHTTP settings template


fodhelper:
Why this values are hardcoded, and why they override my settings from template file
https://github.com/Gozargah/Marzban/blob/325d35d359d6abe15f18338d4dd364010b3b86f4/app/subscription/v2ray.py#L532

M03ED:
These are default values that are provided by xray docs, these values should be set in xray core not in template, that's how it works.

M03ED:
If we ignore the core and just apply the template users gonna face issue if they use 2 or more split http with different settings

PR #1287, opened 2024-09-02
Read SplitHTTP Settings from Template if there is any

fodhelper:
If user add any setting for SplitHTTP in the template, Marzban will use it instead of core or default settings (but by default there is no SplitHTTP settings in the Template file, so will not make any problem for anyone)
Fixes https://github.com/Gozargah/Marzban/issues/1277 and makes users able to fix https://github.com/Gozargah/Marzban/issues/1279 by setting custom settings in the Template file

### DIFF
--- a/app/subscription/v2ray.py
+++ b/app/subscription/v2ray.py
@@ -205,8 +205,14 @@ class V2rayShareLink(str):
 
         elif net == "splithttp":
             # before 1.8.23
-            payload["maxUploadSize"] = sc_max_each_post_bytes
-            payload["maxConcurrentUploads"] = sc_max_concurrent_posts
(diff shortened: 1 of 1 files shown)


M03ED:
this can be good for temporary use but cant solve this 
https://github.com/Gozargah/Marzban/issues/1277#issuecomment-2323106616

M03ED:
As you said it should be a range 
https://github.com/Gozargah/Marzban/issues/1277#issuecomment-2323058459
Wait a few days and I will find a way for this

ImMohammad20000:
This pr hase a merge conflicts can fix it

M03ED:
I have an idea to refactor this part completely but it needs some change in front end

PR #1297, opened 2024-09-03
set `add_as_new_host` default value to false

ImMohammad20000:


xmohammad1:
ye so many times that option ruined my list , good change to be off as defualt

PR #1299, opened 2024-09-04
Prevent zooming on input focus in iOS

MR-MKZ:
When users focus on an input field in iOS, the default behavior is to zoom in to provide a closer view of the text. While this can be helpful in some contexts, it often results in an awkward and jarring experience, especially in responsive web applications. By setting user-scalable to no, we prevent this automatic zooming behavior, ensuring a smoother and more consistent user experience across devices.


MR-MKZ:
wtf :/

PR #1302, opened 2024-09-04
feat(xray-core): new xray features

ImMohammad20000:
- [x] add noise to hosts
- [x] add support for splithttp new variables


app/xray/config.py
```diff
@@ -293,6 +293,8 @@ def _resolve_inbounds(self):
                     settings['scMaxConcurrentPosts'] = net_settings.get('scMaxConcurrentPosts',
                                                                         net_settings.get('maxConcurrentUploads', 100))
                     settings['scMinPostsIntervalMs'] = net_settings.get('scMinPostsIntervalMs', 30)
+                    settings['noSSEHeader'] = net_settings.get('noSSEHeader', False)
```
M03ED, line 296: noSSEHeader:
(Server-only) Do not send the Content-Type: text/event-stream response header. Defaults to false (the header will be sent)
https://xtls.github.io/en/config/transports/splithttp.html#splithttpobject

app/db/models.py
```diff
@@ -214,6 +214,7 @@ class ProxyHost(Base):
     is_disabled = Column(Boolean, nullable=True, default=False)
     mux_enable = Column(Boolean, nullable=False, default=False, server_default='0')
     fragment_setting = Column(String(100), nullable=True)
+    noise_setting = Column(String(100), nullable=True)
```
M03ED, line 217: noise its only for wireguard and currently its not supported

ImMohammad20000: No it work with splithttp h3 too

ImMohammad20000:
> because noise str must be a bytestring, and bytestrings are too long, 100 characters for noise settings is not enough

is 300 character ok?

ImMohammad20000:
> I captured another one and turned it to bytestring, it's 2k characters, i think it's better if you change the limit to 5000

Can you send a picture or type that here?

ImMohammad20000:
I think 1000 characters is enough

ImMohammad20000:
> Some UDP headers are up to 2k, how you think 1000 is enough?

I think it better to remove the character limit

M03ED:
I think we should wait until they complete this feature then add it,

ImMohammad20000:
I will update this to new syntax

ImMohammad20000:
now you can use this syntax to add more than 1 noise
`rand:1-4,10-25&rand:1-4,10-25&base64:7nQBAAABAAAAAAAABnQtcmluZwZtc2VkZ2UDbmV0AAABAAE=,10-25`

ImMohammad20000:
> There is a problem, if you put both maxConcurrency and maxConnections options, Xray will not get started
> Adding them all with value 0 is the problem, if maxConcurrency does not exist in the server config, it must not exist in client config too, creating maxConcurrency with value 0 is not equal to not existence of it
> and as already there is v2ray-settings template, there is no need to add support for xmux!

Is this problem stile exists in the latest version of xray core?

ImMohammad20000:
Please merge this
@M03ED @SaintShit

ImMohammad20000:
I'll add http transport in a separate pr

PR #1335, opened 2024-10-01
fix xray_config.json

iambabyninja:
Currently, there is a major issue because our panel is a very popular solution, and many users are using the demo configuration without changing the parameters, which creates a problem for their own security and the security of the protocol.  
Therefore, I propose leaving the basic configuration and shifting the responsibility for configuration to the user.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -11,7 +11,7 @@ from fastapi_responses import custom_openapi
 
 from config import DOCS, XRAY_SUBSCRIPTION_PATH
 
-__version__ = "0.6.0"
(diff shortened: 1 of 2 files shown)


M03ED (approved):

houshmand-2005:
@iambabyninja 

I think it is better to set the loglevel to `info`
Because IP limit scripts are compatible with loglevel `info` (as far I remember `warning` does not contain client ip log)
Although users have the ability to change this value, many of them do not know that IP limit works with logs and it may cause some confusion.
I don't think there will be much difference in performance or other things

PR #1360, opened 2024-10-12
SECURITY: delete hardcoded admin:password from env

iambabyninja:
Some users use these variables and store their passwords in plain text (not hashed).
Some hosting providers use these variables in their scripts and store passwords in plain text (not hashed).
Some users don’t even change the default values of admin:admin.

To ensure the security of users and their data, this commit allows the creation of administrator accounts only through the CLI interface:

`marzban cli admin create --sudo`


M03ED:
these value will remain in code because it's helpful in development and testing 
sorry but it's there fault, it's not marzban problem

M03ED:
no i don't use it , that will generate dashboard every time and it need to install npm, sometimes you don't need npm

Issue #1384, opened 2024-10-18
ClashMeta Bugs: Mux & HTTPUpgrade


fodhelper:
Problem 1: Mux is disabled in mux template for clash meta (but enabled at host settings), mux will be enabled at clash meta config
Xray server does not support any mux except mux.cool, idk how clashmeta can still connect to my server using h2mux! maybe they have a fallback to disable mux if it's not supported by the server

Problem 2: HTTPUpgrade config does not work with ClashMeta, something must be wrong
I have tested with and without early data, none of them worked

ImMohammad20000:
Pr would be welcome.

mrkonaymyoaung:
same bug here

PR #1386, opened 2024-10-19
Fix Clash Mux

fodhelper:
Fix https://github.com/Gozargah/Marzban/issues/1384


ImMohammad20000:
So It will use what ever in mux template and not change the enable value?

PR #1397, opened 2024-10-24
Fix readme nginx master

fodhelper:


M03ED:
Change branch to dev

ImMohammad20000:
it will update when dev merge into master

PR #1402, opened 2024-10-28
fix zh lang

FreeFlying0:
fix zh lang


FreeFlying0:
The presence of an extra "," at the end of the JSON causes the Chinese translation to not take effect

PR #1416, opened 2024-11-05
feat(tgbot): add Bulk action

ImMohammad20000:
- [x] add create bulk user
- [x] add Bulk User From Template


iamtheted:
Perfect! I appreciate you!
However, this feature is for rich people. 🤣

PR #1463, opened 2024-11-28
fix: incorrect status codes on the subscription page

Immortalety:
Status codes have been changed to comply with the HTTP specification.


ImMohammad20000:
just use raise HTTPException(status_code=404, detail="User not found") and remove other errors

ImMohammad20000:
replace all with that it sending to many information

PR #1468, opened 2024-12-01
add new clash subscription

zhongfq:
I design a subscription system for clash like app, how to use:
1. open the subscription setting in the men.
2. add the proxy with tag, the available inbound of proxy are configed in xray_config.json.

3. open the dialog of user, add subscription domain and the tag(proxy grouped by tag)

4. copy the subscription url

other features:
1. add proxy group

2.add a rule

3. add a ruleset

4. clash setting

5. statistics for users


app/routers/clash.py
```diff
@@ -0,0 +1,1177 @@
...
+        raise HTTPException404("Setting not found")
+
+    crud.update_clash_setting(db, dbsetting, modified)
+
+    logger.info(f"Setting \"{dbsetting.name}\" modified")
+
+    return dbsetting
+
+
+def generate_subscription(db: Session, dbuser: User, user_agent: str):
```
M03ED, line 504: we have a directory for subscription classes and generating them , why all of this cod this in routers ?
routers is just for api requests

app/routers/subscription.py
```diff
@@ -63,6 +63,10 @@ def user_subscription(
                 {"user": user}
             )
         )
+    
+    if len(user.sub_tags) > 0 and len(user.sub_url_prefix) > 0:
+        from app.routers.clash import generate_subscription
```
M03ED, line 68: also your importing from routers in here, if you just follow project structure and put your code in right place there is no need to make unnecessary dependencies

M03ED (changes requested):

app/routers/clash.py
zhongfq: because it has a set of api to add, modify or delete proxy, proxy group, rule, ruleset or settings

app/routers/subscription.py
zhongfq: I want to compat with current subscription of Marzban

ImMohammad20000:
can you dm me in telegram @zhongfq

zhongfq:
> can you dm me in telegram @zhongfq same id at telegram

sorry, I have no telegram

ImMohammad20000:
> > can you dm me in telegram @zhongfq same id at telegram
> 
> sorry, I have no telegram

 can you you add next_plan into the front?
Its in user api

zhongfq:
> > > can you dm me in telegram @zhongfq same id at telegram
> > 
> > 
> > sorry, I have no telegram
> 
> can you you add next_plan to the front? Its in user api

ok, I will try in these days.

SaintShit:
Hello, thanks for you efforts. first check, add_clash_sub migration doesn't run, due to bad revision.
also, it's better to don't inject the clash config to the database, and use templates instead.

zhongfq:
> Hello, thanks for you efforts. first check, add_clash_sub migration doesn't run, due to bad revision. also, it's better to don't inject the clash config to the database, and use templates instead.

image
I want to modify the some clash setting on web. use templates, it mean that i load the setting from templates and write modified setting into the template dynamically?

PR #1483, opened 2024-12-03
fix(api): decorator variable must match the function argument

### LINKED ISSUE #1477: GET /api/user_template/{id} fails
Calling `/api/user_template/{id}` fails with the error: 
```
{"detail":{"template_id":"field required"}} 
```

Indeed `get_user_template` [expects](https://github.com/Gozargah/Marzban/blob/75bcdd98825b4d44a206e78871a9d3cd81af0dfb/app/dependencies.py#L58) a `template_id` argument, while the route [declares](https://github.com/Gozargah/Marzban/blob/75bcdd98825b4d44a206e78871a9d3cd81af0dfb/app/routers/user_template.py#L35) the path parameter as `id`. 

The change was introduced during the routes refactor: https://github.com/Gozargah/Marzban/commit/6918f7fa09e460a123578266920113037637e3c2#diff-ed802903465b7c70a1d639a86b8eeec5304f98e8631a634cf95951326b0b038eR36
Can I submit a PR to fix this ?

vilsbole:
Fix #1477 `/api/user_template` endpoints. 

The routers decorator must match the function argument, so if we use
```def get_user_template(template_id: int, db: Session = Depends(get_db)):``` 

then the decorator should also use `{template_id}` rather than `{id}`

```
@router.get("/user_template/{template_id}", response_model=UserTemplateResponse)
def get_user_template_endpoint(
    dbuser_template: UserTemplateResponse = Depends(get_user_template),
    admin: Admin = Depends(Admin.get_current)):
    """Get User Template information with id"""
    return dbuser_template
```

We could choose either one. This PR makes the arbitrary choice of using `template_id`.

### DIFF
--- a/app/routers/user_template.py
+++ b/app/routers/user_template.py
@@ -32,7 +32,7 @@ def add_user_template(
         raise HTTPException(status_code=409, detail="Template by this name already exists")
 
 
-@router.get("/user_template/{id}", response_model=UserTemplateResponse)
+@router.get("/user_template/{template_id}", response_model=UserTemplateResponse)
 def get_user_template_endpoint(
     dbuser_template: UserTemplateResponse = Depends(get_user_template),
     admin: Admin = Depends(Admin.get_current)):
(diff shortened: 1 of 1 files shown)


ImMohammad20000:
change target branch to dev

vilsbole:
> change target branch to dev

Done ✅ 

Any change you can do a patch release for this ?

PR #1486, opened 2024-12-03
fix: added Marzban's installation commands for MySQL & MariaDB and backup-service commands

iamtheted:


iamtheted:
Special thanks to Alexey

PR #1516, opened 2024-12-15
feat: Add use sni as host

ImMohammad20000:


ImMohammad20000:
> What is it's use case?

when using the random sni feature it make sure to use the same sni as host it is optional

PR #1536, opened 2024-12-22
set two directories index web page for custom template directory

SaberiFardam:
1 template directory = CUSTOM_TEMPLATES_DIRECTORY
2 template directory = NEW_TEMPLATES_DIRECTORY


SaberiFardam:
new directory template web page for set two template index web page

M03ED:
i ask this before 
why do you need this ?
just first one will be used not both of them

M03ED:
as this is second time you open this issue without any explanation and no answer to questions i will close it , if you open another pr like this you will be blocked from gozargah

PR #1552, opened 2024-12-30
Add grouping feature

ImMohammad20000:
# what is new
now you can use groups for better user management
every users have to join to a group

# removes
- remove protocol inbound base system


M03ED:
> is it backward compatible with the old API?
> 

This is gonna be a major update

M03ED:
> If there be a Default group, new version can accept the old type API Requests but ignore the Inbounds list and put the created user to the Default group
> It's not bad to make the API non-strict and backward compatible

In that case there might be some conflict after update with creating users with 3rd party projects, I don't think there be a different 
At the end you need to update your code

ImMohammad20000:
#1713 refactored version

PR #1593, opened 2025-01-10
base configurations for new-frontend

gozarman:
These configurations will be done here
- Migrate to pnpm
- Orval - Tanstack query
- Shadcn (Tailwind) theme configuration based on Figma Design
- New sidebar implementation
- Upgrade React Router to v7


app/dashboard/package.json
```diff
@@ -1,15 +1,19 @@
...
-  "type": "module",
   "scripts": {
-    "dev": "vite --host",
-    "build": "tsc -b && vite build",
-    "lint": "eslint .",
-    "preview": "vite preview"
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "preview": "vite preview",
+    "gen:theme-typings": "chakra-cli tokens ./chakra.config.ts",
```
x0sina: > marzban-ui@0.0.0 gen:theme-typings
> chakra-cli tokens ./chakra.config.ts

'chakra-cli' is not recognized as an internal or external command,
operable program or batch file.
 ELIFECYCLE  Command failed with exit code 1.

app/dashboard/orval.config.js
x0sina, line 1: Rename app/dashboard/orval.config.js to app/dashboard/orval.config.ts 
As It's Typescript project

x0sina (commented):
- Chakra CLI call in Package json
- js file inside Typescript project

x0sina:
Pnpm won't Install
This two packages automaticly
react-hook-form, i18next

x0sina:
If you define BASE_API Like this (Like before)
VITE_BASE_API="http://127.0.0.1:8000/api"
we have api calls like
http://127.0.0.1:8000/api/api/admin/token
but http://127.0.0.1:8000/api/users?limit=10&sort=-created_at this is fine

And If you define like this
VITE_BASE_API="http://127.0.0.1:8000"
We have Correct http://127.0.0.1:8000/api/admin/token api call
but we have wrong users call
but http://127.0.0.1:8000/users?limit=10&sort=-created_at this is fine

PR #1596, opened 2025-01-11
feat: nextPlan -> userTemplate relation

MuhammadAshouri:


app/routers/user.py
```diff
@@ -53,6 +54,13 @@ def add_user(
                 status_code=400,
                 detail=f"Protocol {proxy_type} is disabled on your server",
             )
+            
+    if (new_user.next_plan != None and new_user.next_plan.user_template_id != None):
+        user_template = crud.get_user_template(
```
ImMohammad20000: use get_user_template in dependency file

ImMohammad20000 (changes requested):
can you change branch to target next we are switch to next branch

ImMohammad20000 (approved):

Issue #1603, opened 2025-01-14
Add support for XHTTP: The real upload / download splitting


ImMohammad20000:
https://github.com/XTLS/Xray-core/pull/3955

ImMohammad20000:
#1720

ImMohammad20000:
part 1 is done

PR #1607, opened 2025-01-16
add real upload/download splitting to xhttp

shraiton:
add real upload/download splitting to xhttp by adding downloadSettings in the extra.
the extra part is only for user but for the user to be able to get subscriptions with the extra part we have to use downloadSettings in our configuration in order to be able to send it to the user.

xmux is moved to extra
downloadSettings is moved to extra
keepalive was removed! ( there is a hKeepAlive... option inside xmux ).
subscription tested on v2rayng(android) and v2rayn(windows) and worked fine.
splitting tested and worked fine

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -10,7 +10,7 @@ from fastapi.routing import APIRoute
 
 from config import ALLOWED_ORIGINS, DOCS, XRAY_SUBSCRIPTION_PATH
 
-__version__ = "0.8.2"
+__version__ = "0.8.3"
 
 app = FastAPI(
     title="MarzbanAPI",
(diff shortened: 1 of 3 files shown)


ImMohammad20000:
where is the fallbacks implemented?

shraiton:
> where is the fallbacks implemented?

what do you mean?
splitting connection is already supported by xhttp protocol.
what i just implemented is adding client side code of downloadSettings to the server side ( which is not needed on serverside) for the sole purpose of adding it to the subscription when user update it's sub

ImMohammad20000:
look at the link i provide in #1603

PR #1610, opened 2025-01-16
add real upload/download splitting to xhttp

shraiton:
add real upload/download splitting to xhttp by adding downloadSettings in the extra.
the extra part is only for user but for the user to be able to get subscriptions with the extra part we have to use downloadSettings in our configuration in order to be able to send it to the user.

xmux is moved to extra
downloadSettings is moved to extra
keepalive was removed! ( there is a hKeepAlive... option inside xmux ).
subscription tested on v2rayng(android) and v2rayn(windows) and worked fine.
splitting tested and worked fine


M03ED:
your trying to make a pr from master branch dev, this is not allowed and you already have another open pr

PR #1616, opened 2025-01-19
fix: dynamically set https for VITE_BASE_API when SSL is enabled

Joursoir:
If DEBUG is turned on and SSL mode is enabled, API requests will be sent to the http address, which is unavailable in SSL mode, making it impossible to reach the API.


Joursoir:
@ImMohammad20000 @gozarman hello! Do you need any extra comments on this patch? I will be happy to provide whatever you need

Joursoir:
@ImMohammad20000 @gozarman ping

ImMohammad20000:
Hi @gozarman need to approve this change

Issue #1641, opened 2025-02-09
New users can't use `Vision` if `network` is `raw` until Xray restarted.


Akiyamov:
**Describe the bug**
If new user is added and `network` is `raw` new users won't be able to use `Vision` until restart of Xray.

**To Reproduce**
1. Use [this config](https://iambabyninja.github.io/xray-config/?c=G+AGAJwHto1sm0yL0afgrjNgnX0lXn+xARroBhHp6pssCjoJotDlf42tIRRCzKy8jyeTRFNN8HdxDZWhBDvJ4UI8HYP5TkIZ7pWf/0GH8xpb0oFKIYsh0fw4PnVNQ9tTXMp3rK8IE9jo7TFdY6i/8ZB0aaT/u0z8H3JnUNHYy2mIJjdZIYmOGN3iee5IqrDU973iy44oJu5qDkpo7FR19mxes1jOb99Li5drmXvTEJx1hxrPxQu3wMRxD7fpG1PCR46QQEZRHG3vsRv6vJa34RxmhYr67vHNgmElAOKMyu89hIOGTjXZTpkk4IVBGYobzZNF5gXZHQqWQfidD3sF0yVKpbheegtppe3+t7bBrvQjaTEt7u522xJt5lfCOH8XW7WMY7zF5DkjCgI7Y2V2OV9N99gUnB8s8azInLard8VvkQ9ozg71rOdmfOS99yQp2VUh2McqsArAo4/DHYXfHynK+weiRPM7FcU+GPGmZ08IfDvFCcp+4AxN5iNsxprL6K6bfwpDdV966YtHLZZWWEWU8pyAzVhx7lyEBvFkgfdH3CyQELIsSm77eJxvZyG44SAEc8py8VV3zhw31yTvreLz6Y6rDOVvL07BJxYw1CYVQ26yQQvJZHc+SqNOqY0A9svRY9rkfW3lc1fiTAGNCBdV3AZi+vyghe8=) as example
2. Connect to this client using `Vision`
3. Get `2025/02/09 19:04:44 [Warning] [3974872192] app/proxyman/inbound: connection ends > proxy/vless/inbound: account 40c147e3-b6d6-43f3-969a-51dc3bcb63de is not able to use the flow xtls-rprx-vision`
4. Restart Xray 
5. Connect with `Vision` again without problem

**Expected behavior**
We connect with `Vision` from start 

**Machine details (please complete the following information):**
[...]

PR #1642, opened 2025-02-09
Fix issue #1641

Akiyamov:
Seems like we need to add `raw` just like [here](https://github.com/Gozargah/Marzban/blob/7f396db3e703d71a28060bc9ce4a532ec64cb1f4/app/xray/config.py#L417)


ImMohammad20000:
change target branch to dev

Akiyamov:
I did, didn't i? Or i should change to `dev` in my repo too?

ImMohammad20000:
no its ok thanks

Issue #1638, opened 2025-02-09
Ports exposed via docker port doesn't accessible


fractalshine:
**Describe the bug**
Panel doesn't accessible if ports mapped via 'port' directive

**To Reproduce**
Update from 0.6.0 to latest tag (0.8.4 for now). Marzban panel behind nginx proxy with ssl. After upgrade I've got 502 in browser, trying to connect through nginx and "The connection was reset" when port mapped though ssh. On my vps can't curl even curl to mapped localhost container (127.0.0.1:9000) - "curl: (56) Recv failure: Connection reset by peer". The only way to connect is - from inside container.
Part of my docker-compose.yml

```
services:
  marzban:
    image: gozargah/marzban:latest
    restart: always
    env_file: .env
      #network_mode: host
    ports:
      - "127.0.0.1:9000:8000"
      - "127.0.0.1:8090:8080"
    volumes:
      - /var/lib/marzban-tests:/var/lib/marzban
    depends_on:
      - mysql
   ```

**Additional context**
I assume [this](https://github.com/Gozargah/Marzban/issues/1400#issuecomment-2441082294) code has to be in upstream.

PR #1645, opened 2025-02-10
Allow connecting from whole LAN, not only localhost

### LINKED ISSUE #1400: 502 Bad Gateway after update from v0.6.0 to v0.7.0
I use docker and set up docker compose config with Marzban and Traefik Proxy. On Marzban version 0.6.0 everything works, but after upgrading to version 0.7.0 I get an error
I also used Nginx, which also returns an error
When trying to open the image URL directly (http://172.18.0.4:8000/), I got HTTP ERROR 502

**Machine details (please complete the following information):**
 - OS: Ubuntu 22.04
- Docker Compose version v2.27.0

Akiyamov:
This PR should close #1400 and #1638. As i can see in first issue was done but code wasn't merged.

I'm not using function `is_global` from `ipaddress` library since it counts `0.0.0.0` as private.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,3 +1,4 @@
+import ipaddress
 import click
 import logging
 import os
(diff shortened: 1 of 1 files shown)


M03ED:
i write a code for this few month ago but i forgot to merge it 
i add this in this commit [`c47429f34aa6640892273d604e7b864a2a167081`](https://github.com/Gozargah/Marzban/commit/c47429f34aa6640892273d604e7b864a2a167081)

Akiyamov:
I think you need to add small check for `0.0.0.0` address otherwise it would allow user to bind without certs.
```
>>> ipaddress.ip_address("0.0.0.0").is_private
True
>>> ipaddress.ip_address("10.5.0.1").is_private
True
```

M03ED:
> I think you need to add small check for `0.0.0.0` address otherwise it would allow user to bind without certs.
> 
> ```
> >>> ipaddress.ip_address("0.0.0.0").is_private
> True
> >>> ipaddress.ip_address("10.5.0.1").is_private
> True
> ```

good point

Akiyamov:
@M03ED [In this commit](https://github.com/Gozargah/Marzban/commit/f5270211c1cc565ce2d01ecddb539856c376bbfa) code compares ip class and string, so it would be false. I think it should be like this
```
>>> ip = ipaddress.ip_address("0.0.0.0")
>>> ip == "0.0.0.0"
False
>>> ip == ipaddress.ip_address("0.0.0.0")
True
```

PR #1714, opened 2025-03-17
feat(notification): Added notification for host operations

Atomic8oy:
The Telegram part is not tested.


app/models/host.py
```diff
@@ -196,15 +196,15 @@ class BaseHost(BaseModel):
     security: ProxyHostSecurity = ProxyHostSecurity.inbound_default
     alpn: ProxyHostALPN = ProxyHostALPN.none
     fingerprint: ProxyHostFingerprint = ProxyHostFingerprint.none
-    allowinsecure: bool | None = None
-    is_disabled: bool | None = None
+    allowinsecure: bool = False
```
ImMohammad20000: this change break hosts

  {'type': 'bool_type', 'loc': ('response', 22, 'allowinsecure'), 'msg': 'Input should be a valid boolean', 'input': None}

ImMohammad20000 (changes requested):

PR #1726, opened 2025-03-24
Optional field data_limit and expire remove from dict in get_subscrip…

mangustyura:
Optional field data_limit and expire remove from dict in get_subscription_user_info if field None

Adding optional fields with a value of 0 to the output breaks the logic in cases where the user hasn’t configured their panel at all and hasn't provided the necessary parameters, as well as in cases where they explicitly specify that the traffic is unlimited.
For users who haven't configured anything, we don’t want to show any metadata at all.
And for those who have configured it, if they've used 0 traffic and have an unlimited plan, we want to display that the traffic is unlimited.
To distinguish between these two cases, the total field should not be sent if it doesn't exist.


ImMohammad20000:
In some apps like hiddify not having some parameters break the ui interface so we need to have that part in headers to prevent that

PR #1827, opened 2025-05-27
Update .env.example

korn3r:
fixed typo in UVICORN_UDS parameter


ImMohammad20000:
this version is not supported by us anymore

PR #1844, opened 2025-06-28
feat: Add dashboard to front-end

milad-mamandi:


milad-mamandi:
photo_5976367115038475832_w

M03ED:
why is this with mock data ?
shouldn't we use real api ?

M03ED:
and why `Marzban.code-workspace` is in gitignore file ?

milad-mamandi:
> why is this with mock data ? shouldn't we use real api ?

from what I've looked at the API currently doesn't provide the necessary data (no info on traffic usage for each admin, only total traffic, info on users are also not completed, no api for expired/limited/onhold/deactivated users)
i mentioned that i need to be in touch with the backend developer to get more info

milad-mamandi:
> and why `Marzban.code-workspace` is in gitignore file ?

to be honest idk, I only added node_modules

M03ED:
> > why is this with mock data ? shouldn't we use real api ?
> 
> from what I've looked at the API currently doesn't provide the necessary data (no info on traffic usage for each admin, only total traffic, info on users are also not completed, no api for expired/limited/onhold/deactivated users) i mentioned that i need to be in touch with the backend developer to get more info

all of them are available already in api

x0sina:
Hey Thanks for your pr please resolve the conflicts remove the changes on gitignore its already ignoring node modules on main directory and your ignoring workspace to use real data the data stored in src/service/api/index.ts file live from Marzban Swagger UI

M03ED:
i think your using useGetSystemStats twice, once for showing cpu and memory , next one for user statistic but one request should be enough.
can we hide admin username for non sudo admins ?
they can't send this parameter or even they do not gonna works

milad-mamandi:
> i think your using useGetSystemStats twice, once for showing cpu and memory , next one for user statistic but one request should be enough. can we hide admin username for non sudo admins ? they can't send this parameter or even they do not gonna works

got a bit too complicated for no reason, will refactor later, but should work just fine for now

PR #1849, opened 2025-06-29
Fix ObjectDeletedError in review job

xmohammad1:
## Summary
- handle `ObjectDeletedError` in `review` job to avoid crashes when users are removed

## Testing
- `python -m py_compile $(git ls-files '*.py')`

## Summary
- avoid out of range errors by storing expire timestamps as `BigInteger`
- add Alembic migration to update `users.expire` and `next_plans.expire`

## Testing
- `python -m py_compile app/db/models.py app/db/migrations/versions/9953c54cf26d_users_expire_bigint.py`

### DIFF
--- a/app/db/migrations/versions/9953c54cf26d_users_expire_bigint.py
+++ b/app/db/migrations/versions/9953c54cf26d_users_expire_bigint.py
@@ -1,0 +1,28 @@
+"""use BigInteger for expire columns
+
+Revision ID: 9953c54cf26d
+Revises: 2b231de97dc3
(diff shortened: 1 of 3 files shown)


M03ED:
Thanks for you pr.
But I should tell you expire have been changed to datetime in `v1`
Also next plan usage have been changed.
But for your change in review function I should say it's good change for `v0` but these function have been changed completely in `v1` and I don't think these situations happens again unless you do some raw query usage directly on database
If you want to submit new pr you should submit it for `next` branch so we can approve it.
Have a good day.

PR #1852, opened 2025-06-30
feat: onHoldTimeOut added to user create modal

milad-mamandi:
feat: onHoldTimeOut added to user create modal
fix: expiry date would show incorrect value when switching status


x0sina:
please add other languages translations

x0sina:
also translations not needed to pass with props its static

x0sina (approved):
Nice Work Approved

PR #1853, opened 2025-06-30
feat(admin): display all user statuses and recent online users in system info

Joursoir:
Add all user statuses and "online in last 24h" to the system info screen. Now the Telegram bot shows the same info as the API.

This matches what the `/system` API already does (added in #1445).


ImMohammad20000:
We are not add new features on this branch anymore just fixes 
If you want to help checkout the next branch

PR #1861, opened 2025-07-07
Enhance Telegram notifications formatting and trigger on user add

AmirKenzo:


M03ED:
notification is just for you to notice you some changes happens not show all the details
also in some cases this might exceed length limit

M03ED (commented):
also * is not gonna work on Hashtags

AmirKenzo:
> also * is not gonna work on Hashtags

I tested it: with the * the log is sent without any issues, but without the * it doesn’t get sent.

M03ED:
It will be sent but is useless

AmirKenzo:
> It will be sent but is useless 

Well, we’re currently using Markdown version 1, and double asterisks don’t work—only a single asterisk should be used, but most of the code uses two.
2025-07-07  23-21-25_Telegram

PR #1865, opened 2025-07-10
Make the server response adhere to the XTLS Subscription Standards

jomertix:
Hello. Please update Marzban to be sure the panel fully comply XTLS Subscription Standards. Since expire=0 is a valid timestamp, the server should not return this subfield in the subscription-userinfo header if the subscription does not have an expiration date.

See https://github.com/XTLS/Xray-core/discussions/4877


M03ED:
first of all making pr directly on master branch is not allowed
second this version of marzban is not gonna get any update until v1 become stable
if you have any change you want to submit i suggest to do it for `next` branch

PR #1893, opened 2025-07-20
General Settings

gozarman:
- **feat: general settings section implemented**
- **feat: autofill `vless flow` and `shadowsocks method` based on general settings for new users**


dashboard/src/pages/_dashboard.settings.general.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+export default function General() {
+  const { t } = useTranslation()
+  const { settings, isLoading, error, updateSettings, isSaving } = useSettingsContext()
+
+  const form = useForm<GeneralSettingsForm>({
+    resolver: zodResolver(generalSettingsSchema),
+    defaultValues: {
+      default_flow: '',
+      default_method: '',
```
ImMohammad20000, line 31: set the default method to chacha20-ietf-poly1305

ImMohammad20000 (changes requested):

x0sina:
Great Job
There was just one problem that I fixed 
it was using /settings instead of /settings/general

PR #1900, opened 2025-07-25
Advance search

Normal-DV:
feat: implement Advanced Search dialog with filter support and update translations

- Added Advanced Search as a dialog component
- Applied selected filters to user list on submission
- Updated localization files and translations for all supported languages
- Add badge for selected group and admin in advance search modal

image
image
image
image


x0sina (approved):
Everything ok and there is no conflict with search and filters logic
I also added clear all filters and filters count badge
and removed duplicate translations

Normal-DV:
Thanks for the review and changes!

PR #1902, opened 2025-07-28
Fix UI bugs

Normal-DV:
Added confirm modal for activating next plan

Improved username generator with better randomness

Fixed bugs in user modal (showing 0, wrong expire time)

Fixed username prefix/suffix edit issue

Updated labels and adjusted layout in next plan forms


x0sina (approved):
This pr fixed some bugs in user template modal and usermodal
also have a confirmation modal to activate next plan 
All Aprroved Thx

PR #1909, opened 2025-07-31
fix: UI bugs

Normal-DV:
this pr is for fixing some ui bugs 

fix : search input and buttons color in user and admin table
fix : previous and next in admin pagination 
fix : show validation in user modal after user touch or submit

image
image


x0sina (approved):
Approved

PR #1917, opened 2025-08-07
feat: Implement host-based groups functionality and access on subscription

BemoBit:
This commit introduces the ability to associate ProxyHosts with Groups, similar to the existing inbound_tags functionality. Key changes include:

- Added hosts_groups_association table in database models
- Created migration for the new association table
- Extended Group model to include host_ids field
- Updated Group validation to require at least one inbound or host selection
- Modified Group CRUD operations to handle host associations
- Updated GroupModal UI to include host selection alongside inbound tags
- Ability to display selected hosts from the user's group in the output subscription list
- Adjusted sidebar and menu labels for hosts section
- Updated dashboard default values to include host_ids
- Enhancing host translations across various languages

This change allows administrators to create groups based on specific hosts rather than just inbound tags, providing more granular control over user access.


ImMohammad20000:
first of all test dose not pass
also having group based hosts not a good idea i didn`t read your code prorly but a consept for grouping hosts have security risks

PR #1925, opened 2025-08-13
fix: add translation for Sort by expire

dzaytsev91:
Add ru translation for user table, and add translation function for "Sort by expire" for all languages


ImMohammad20000:
This version is not supported by us anymore checkout next branch

PR #1963, opened 2025-11-27
feat(config): add fallback destination mapping

4erdenko:
## Context
`xray_config.json` may contain “virtual” inbounds that listen on pseudo addresses `@...` and must inherit port and Reality settings from a parent inbound (the one that owns a `fallbacks.dest`). Stock Marzban could inherit only from a single global fallback (`XRAY_FALLBACKS_INBOUND_TAG`), so such schemes failed: inbounds without their own `port`/`realitySettings` were dropped or produced TCP‑Reality instead of XHTTP‑Reality.

## What changed
- Added a map `fallback_dest -> inbound` in `app/xray/config.py`, built from `settings.fallbacks.dest`.
- When parsing an inbound without `port`, we now first look up a parent by `listen` (e.g., `@child-xhttp` → fallback with `dest: "@child-xhttp"`); only if none is found do we use the old `XRAY_FALLBACKS_INBOUND_TAG`.
- `security`/`realitySettings` are also inherited from the found parent; the inbound keeps its own `network` (xhttp/ws/…).
- Backward compatibility preserved: the env var still works as before; new branches are just additionally supported.

## Why it’s useful
- Supports multiple fallback branches at once (e.g., TCP Reality parent + XHTTP Reality child) without duplicating ports/keys.
- Generates correct client links like `vless://…?type=xhttp&security=reality&pbk=…&sid=…` for inbounds that don’t listen on a real port and sit on an `@alias`.
- Lets you combine WS/XHTTP/Reality chains in `xray_config.json` without sacrificing any of them.

## Safety
- Inheritance logic applies only when an inbound has no `port` and `listen` starts with `@`; others are untouched.
- Priority kept: explicit `XRAY_FALLBACKS_INBOUND_TAG` still wins if present.
- If no suitable parent is found, behavior doesn’t worsen: validation still complains about a missing `port`.

## What becomes possible
- A setup with “parent Reality on TCP, port N” + “child XHTTP on `@alias`” yields an XHTTP+Reality link with port/PBK/SID inherited automatically.
- WS branches can rely on one fallback while XHTTP branches rely on another, simultaneously.

## Mini config example
```json
{
  "inbounds": [
    {
      "tag": "PARENT-REALITY",
      "port": 1443,
      "protocol": "vless",
      "settings": { "decryption": "none", "fallbacks": [ { "dest": "@child-xhttp" } ] },
      "streamSettings": {
        "network": "tcp",
        "security": "reality",
        "realitySettings": { "publicKey": "PBK", "shortIds": ["abcd1234"] }
      }
    },
    {
      "tag": "CHILD-XHTTP",
      "listen": "@child-xhttp",
      "protocol": "vless",
      "settings": { "decryption": "none" },
      "streamSettings": { "network": "xhttp", "xhttpSettings": { "path": "/home" } }
    }
  ]
}
```
Result: `CHILD-XHTTP` inherits port and `realitySettings` from `PARENT-REALITY`; the generated link contains `type=xhttp&security=reality&pbk=PBK&sid=abcd1234`.

For reference, a typical XHTTP+Reality setup can be seen here: https://github.com/chika0801/Xray-examples/tree/main/VLESS-XHTTP-REALITY/steal_oneself

## What to verify
- An inbound with `listen: "@alias"` and no `port` receives port/TLS/PBK/SID from the parent whose `fallbacks.dest == "@alias"`.
- Generated subscription for such an inbound includes `security=reality`, `pbk`, `sid`, correct port, and preserves its `network` (ws/xhttp/etc.).
- Legacy configs using only `XRAY_FALLBACKS_INBOUND_TAG` still work unchanged.

## Environment/frontend notes
- No frontend changes are required (parsing is backend-only).
- `XRAY_FALLBACKS_INBOUND_TAG` can be left empty if inheritance is driven by `fallbacks.dest`.

### DIFF
--- a/app/xray/config.py
+++ b/app/xray/config.py
@@ -57,6 +57,8 @@ class XRayConfig(dict):
         self.inbounds_by_protocol = {}
         self.inbounds_by_tag = {}
         self._fallbacks_inbound = self.get_inbound(XRAY_FALLBACKS_INBOUND_TAG)
+        # map fallback destination (e.g. "@xhttp-glance") -> inbound that owns the fallback
+        self._fallbacks_by_dest = {}
         self._resolve_inbounds()
 
         self._apply_api()
@@ -141,6 +143,13 @@ class XRayConfig(dict):
                 raise ValueError("all outbounds must have a unique tag")
 
     def _resolve_inbounds(self):
+        # collect fallback destinations to their parent inbound for later inheritance
+        for inbound in self['inbounds']:
+            for fb in inbound.get('settings', {}).get('fallbacks', []):
+                dest = fb.get('dest')
+                if isinstance(dest, str) and dest.startswith('@'):
+                    self._fallbacks_by_dest[dest] = inbound
+
         for inbound in self['inbounds']:
             if not inbound['protocol'] in ProxyTypes._value2member_map_:
                 continue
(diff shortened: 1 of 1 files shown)


4erdenko:
@codex review

Issue #672, opened 2023-12-02
Problem with getting logs from Xray core When we save logs


houshmand-2005:
Hi, The problem is that when we save logs, logs are no longer displayed in the panel and API
<hr>

After change config(xray_config.json) like this
```json
"log": {
    "loglevel": "info",
    "access": "access.log",
    "error": "error.log"
  },
```
This is what Api and Panel shows:
```
Xray 1.8.4 (Xray, Penetrates Everything.) Custom (go1.21.0 linux/amd64)

A unified platform for anti-censorship.
2023/11/22 16:59:16 [Info] infra/conf/serial: Reading config: stdin:

```
And this is the contents of  `access.log` file (this is right and ok):
```
2023/11/22 17:01:32 192.168.1.33:11421 accepted tcp192.168.1.36:443 [VLESS_INBOUND -> proxy] email: 1.Test
2023/11/22 17:01:36 192.168.1.33:11424 accepted tcp192.168.1.36:443 [VLESS_INBOUND -> proxy] email: 1.Test
2023/11/22 17:01:37 192.168.1.33:11430 accepted tcp192.168.1.36:443 [VLESS_INBOUND -> proxy] email: 1.Test
```

As I can understand when with Xray we save the logs Xray doesn't show them on stdout, so marzban can't catch them.
This is [Marzban code](https://github.com/Gozargah/Marzban/blob/7dfbe1dc7e4864c27360e47f4d7567db4cdc65e4/app/xray/core.py#L68C1-L71C1) to get logs:

```python
    ...
    def capture_only():
        while self.process:
            output = self.process.stdout.readline()
            if output:
                output = output.strip()
                ...
```

so marzban just read the stdout and when Xray is saving log file, it doesn't send logs into stdout.

I found this out when someone reported it in my script   https://github.com/houshmand-2005/V2IpLimit/issues/32

And if we remove the save location to just show the logs like this:
```json
"log": {
    "loglevel": "info"
  },
```
Again the logs are sent without any problem.

<hr>

Do you think something can be done for this?

omid-sadeghi:
با سلام
[...]

PR #2092, opened 2026-09-07
fix(xray): drain stderr through log capture

zoomerland:
## Summary

- merge Xray's stderr into the stdout stream already consumed by Marzban's log-capture thread
- prevent an unread stderr pipe from filling and blocking Xray
- add isolated regression coverage for log capture and a real child process writing more than 2 MiB to stderr

## Problem

`XRayCore.start()` currently creates separate `stdout=PIPE` and `stderr=PIPE` streams, while `__capture_process_logs()` reads only `process.stdout`. Once Xray writes enough data to fd 2, the stderr pipe fills and the writer blocks.

This was reproduced on Marzban 0.8.4 with Xray 26.3.27. A goroutine profile showed an Xray `net/http` error logger blocked in `syscall.write(fd=2)` for hundreds of minutes, with hundreds of other goroutines waiting on the same standard-library logger mutex. File descriptors and `CLOSE_WAIT` sockets then accumulated until service degradation.

## Fix

Use `stderr=subprocess.STDOUT`. The existing capture thread now continuously drains both output streams through the same bounded log buffers. No process-lifecycle, configuration, subscription, or public API behavior is changed.

## Tests

```text
python -W error::ResourceWarning -m unittest discover -s tests -v
Ran 2 tests ... OK
```

The same suite passes on Windows and Linux/WSL. The real-pipe regression writes over 2 MiB to stderr before emitting stderr/stdout markers; the patched process completes without blocking. Against the original `dev` implementation, the test reaches its bounded 10-second timeout as expected.

Related to #672 and #911.

### DIFF
--- a/app/xray/core.py
+++ b/app/xray/core.py
@@ -120,7 +120,9 @@ class XRayCore:
             cmd,
             env=self._env,
             stdin=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            # Xray writes runtime logs to stderr, so merge it into the
+            # stdout stream already drained by the log capture thread.
+            stderr=subprocess.STDOUT,
             stdout=subprocess.PIPE,
             universal_newlines=True
         )
--- a/tests/test_xray_core.py
+++ b/tests/test_xray_core.py
@@ -1,0 +1,173 @@
+import importlib.util
+import io
+import pathlib
+import subprocess
+import sys
+import textwrap
+import time
+import types
+import unittest
+from unittest import mock
+
+
+CORE_PATH = pathlib.Path(__file__).parents[1] / "app" / "xray" / "core.py"
+REAL_POPEN = subprocess.Popen
+
+
+def load_core_module():
+    app_module = types.ModuleType("app")
+    app_module.__path__ = []
+    app_module.logger = mock.Mock()
+
+    xray_module = types.ModuleType("app.xray")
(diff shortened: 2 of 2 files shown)


zoomerland:
## Production validation update

The patch has now been running in production for 71.2 hours, covering 4,121 valid one-minute samples and three nearly complete Xray process cycles separated only by scheduled daily host reboots.

Results compared with the 21.3-hour pre-fix baseline:

- no resource-guard warnings or unscheduled Xray restarts after the fix;
- no OOM events or critical Marzban log patterns;
- `CLOSE_WAIT` median: 10 -> 0, p95: 95 -> 12, maximum: 147 -> 31;
- FD p95: 501 -> 174;
- swap remained 0 throughout;
- each post-fix Xray process remained healthy for approximately 23.4-23.9 hours until the scheduled host reboot;
- the writable-layer hotfix survived two ordinary host reboots, and stderr/stdout remained connected to the same actively drained pipe.

The post-fix window still included substantial transient load: FD briefly reached 781, but recovered instead of accumulating. At the latest check, the panel returned HTTP 200, proxy ports were listening, `CLOSE_WAIT` was 0, and the Xray process had been running for 23 hours 24 minutes.

This is strong production evidence that draining stderr prevents the observed pipe-blocking failure mode. It does not claim that every superficially similar Marzban outage has this same cause.

